PR #3, opened 2019-06-27
Initial commit

chuckha:


chuckha:
this is two commands

```
kubebuilder init --domain bootstrap.k8s.io
kubebuilder create api --group kubeadm --version v1alpha1 --kind KubeadmBootstrapConfig
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Build the manager binary
+FROM golang:1.12.5 as builder
```
detiber: Should this be bumped to 1.12.6?

Makefile
```diff
@@ -0,0 +1,64 @@
+
+# Image URL to use all building/pushing image targets
+IMG ?= controller:latest
+# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
+CRD_OPTIONS ?= "crd:trivialVersions=true"
```
detiber: It might be good to go ahead and not use trivial versions to allow easier upgrades later.

PROJECT
```diff
@@ -0,0 +1,7 @@
+version: "2"
+domain: bootstrap.k8s.io
```
detiber: This should be something under `sigs.k8s.io` and should likely be more easily traceable back to the cluster-api subproject.

chuckha:
/hold

not sure this is right

hack/boilerplate.go.txt
```diff
@@ -0,0 +1,14 @@
+/*
```
detiber: This is missing the copyright

chuckha: the boilerplate file is missing the copyright?

Dockerfile
chuckha: yeah, i'll do that on a regen, i haven't read through this generated stuff yet

Makefile
chuckha: i don't even know what that is, i still have to read through this stuff 😆 (hence the hold)

PROJECT
chuckha: bootstrap.clusterapi.sigs.k8s.io ?

detiber: I think @vincepri used `cluster.sigs.k8s.io`, so maybe `bootstrap.cluster.sigs.k8s.io`?

hack/boilerplate.go.txt
detiber: Yes, see: https://github.com/kubernetes-sigs/cluster-api/blob/master/hack/boilerplate.go.txt

Makefile
vincepri: Unless you have multiple versions, that flag has to be used for making it work with controller-tools v2

detiber: https://book.kubebuilder.io/print.html#multiple-versions

PROJECT
vincepri: `bootstrap.cluster.sigs.k8s.io`

Makefile
```diff
@@ -0,0 +1,64 @@
...
+
+# Push the docker image
+docker-push:
+	docker push ${IMG}
+
+# find or download controller-gen
+# download controller-gen if necessary
+controller-gen:
+ifeq (, $(shell which controller-gen))
+	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.0-beta.2
```
vincepri: You shouldn't run this in the same repository, or it'll add the dependencies to go.mod, unless I missed something

vincepri: In CAPI, we build and reference vendor/

Makefile
```diff
@@ -0,0 +1,64 @@
...
+CONTROLLER_GEN=$(shell go env GOPATH)/bin/controller-gen
```
vincepri: GOPATH shouldn't be used with go modules

Makefile
```diff
@@ -0,0 +1,64 @@
...
+else
+CONTROLLER_GEN=$(shell which controller-gen)
```
vincepri: I'd generally advise building utilities from vendor and putting them in a bin/ folder

api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,35 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "kubeadm.bootstrap.k8s.io", Version: "v1alpha1"}
```
vincepri: The group should be `bootstrap.cluster.sigs.k8s.io`

hack/boilerplate.go.txt
vincepri: This is new from k/k and adapted https://github.com/kubernetes-sigs/cluster-api/tree/master/hack/boilerplate

PROJECT
chuckha: works for me

hack/boilerplate.go.txt
chuckha: ah right, got it

neolit123:
/retest
verify presubmit should be up

sethp-nr:
@vincepri I think the Makefile is stock from the kubebuilder upstream, so whatever changes we want to make we should consider sending along too.

That said, I'm not sure we have a `vendor` in the brave new go-mod world. So far the guidance I've found for versioned tool dependencies is https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module which suggests importing the command (!) in a `tools` package. I couldn't really get that working well for me, but since `go get` and `go env GOPATH` are module-aware the provided Makefile does "work", even if it makes me uncomfortable.

chuckha:
@vincepri your comments are valid but as @sethp-nr pointed out this is all generated code from kubebuilder. I'll head over there and see what master is looking like and provide some feedback if necessary

Makefile
chuckha: This seems pretty minor, we can run `go mod tidy` after this command (embed it in the makefile target even) and it will clean up go.mod and go.sum

Makefile
chuckha: i think what you're getting at is that this makefile is using go to manage dependencies. We should not be using go to manage dependencies and in place of this makefile target we should be doing a check to make sure we have controller-gen in the path. I agree that scope creeped in this Makefile.

chuckha: actually this target already does that. It looks in the path then tries to go get which is a nice experience but could be a bit weird.

Overall I don't think this is exceptional enough to change from the boilerplate.

Makefile
chuckha: vendor is being phased out in favor of a go proxy which we can leverage using a script like this instead of a direct go get:

```
#!/usr/bin/env bash
VERSION=${VERSION:-0.2.0-beta.2}

# get out of the module
cd /tmp
GO111MODULE=on go get "sigs.k8s.io/controller-tools/cmd/controller-gen@v${VERSION}"
cd -
```

Here is my thinking that led me down this path:

* we can't use `go get` within a go module (as you pointed out, it modifies go.mod/go.sum)
* controller-runtime doesn't release binaries so we must build from source if we don't have `controller-gen` installed (this makes using go get VERY appealing)
* vendor is being phased out in favor a proxy (don't vendor the source to build from vendor)
* outside of the GOPATH we can't use the version specification (must be outside of any module but still use go get with module mode)

which is how this script ended up with getting outside of the module but still using go get (in module mode) to install the dependency

vincepri:
@chuckha Sounds like a good plan.

@sethp-nr Makes sense, we can (should?) still vendor with `go mod vendor`, as we do upstream. It's been useful to include the vendor folder for a few reasons: backwards compatibility with older Go versions, shorter CI time using `GOFLAGS="-mod=vendor"`, and tooling

chuckha:
closing this, thank you for the reviews everyone!

PR #8, opened 2019-07-02
Fix verify-all

chuckha:

/assign @neolit123

### DIFF
--- a/hack/verify-all.sh
+++ b/hack/verify-all.sh
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set -o errexit
 set -o nounset
 set -o pipefail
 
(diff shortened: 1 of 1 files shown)


chuckha:
/hold

no that's not right either.

chuckha:
/hold cancel

fixed

hack/verify-all.sh
```diff
@@ -37,64 +37,64 @@ res=0
 
 if [[ "${VERIFY_WHITESPACE:-true}" == "true" ]]; then
   echo "[*] Verifying whitespace..."
-  hack/verify-whitespace.sh || res=1
-  failure $res "verify-whitespace.sh"
+  hack/verify-whitespace.sh
+  failure $? "verify-whitespace.sh"
```
neolit123: hm, maybe i'm missing something. can you explain what the fix is here?
each `verify-*.sh` has a `|| res=1` which is fine and `failure()` only notifies in this case.
then near EOF `res=1/0` triggers an overall summary and the exit code matches `res`.

neolit123: ok, i see.
`$res` can leak from previously failed `verify-` calls.

PR #10, opened 2019-07-03
Add GitHub issue and PR templates

vincepri:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,25 @@
...
+**What this PR does / why we need it**:
+
+**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
+Fixes #
+
+**Special notes for your reviewer**:
+
+_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._
+
+**Release note**:
```
neolit123: i don't think we have the release-note plugin enabled for this repo (i.e. `do-not-merge/release-note-label-needed` will not be applied to PRs without release notes).

also i guess a question is - do we need release notes @chuckha ?

neolit123: hm, looks like this was added here already:
https://github.com/kubernetes-sigs/cluster-api-provider-docker/blob/master/.github/PULL_REQUEST_TEMPLATE.md

chuckha: They are just kinda here by default, easy enough to ignore and there are even instructions for NONE so I think this is fine for consistency

PR #25, opened 2019-07-12
Adds some manager code that does nothing

chuckha:
This PR adds scaffolding and some fixes to make testing a little bit nicer

The code doesn't do anything very special, I'm not even sure if it's right

**Release note**:
```release-note
None
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,13 @@
 
-# Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
+# A release does not need to define this
+MANAGER_IMAGE_NAME ?= cluster-api-bootstrap-provider-kubeadm
+MANAGER_IMAGE_TAG ?= dev
+PULL_POLICY ?= Always
+
+# Used in docker-* targets.
+MANAGER_IMAGE ?= $(REGISTRY)/$(MANAGER_IMAGE_NAME):$(MANAGER_IMAGE_TAG)
+
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
 CRD_OPTIONS ?= "crd:trivialVersions=true"
 
@@ -45,13 +52,13 @@ generate: controller-gen
 
 # Build the docker image
 docker-build: test
-	docker build . -t ${IMG}
+	docker build . -t ${MANAGER_IMAGE}
 	@echo "updating kustomize image patch file for manager resource"
-	sed -i'' -e 's@image: .*@image: '"${IMG}"'@' ./config/default/manager_image_patch.yaml
+	sed -i'' -e 's@image: .*@image: '"${MANAGER_IMAGE}"'@' ./config/default/manager_image_patch.yaml
 
 # Push the docker image
 docker-push:
-	docker push ${IMG}
+	docker push ${MANAGER_IMAGE}
 
 # find or download controller-gen
 # download controller-gen if necessary
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,12 +1,12 @@
 # Adds namespace to all resources.
-namespace: cluster-api-bootstrap-provider-kubeadm-system
+namespace: cabpk-system
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: cluster-api-bootstrap-provider-kubeadm-
+namePrefix: cabpk-
 
 # Labels to add to all resources and selectors.
 #commonLabels:
--- a/controllers/kubeadmbootstrapconfig_controller.go
+++ b/controllers/kubeadmbootstrapconfig_controller.go
@@ -21,14 +21,10 @@ import (
 	"fmt"
 
 	"github.com/go-logr/logr"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	kubeadmv1alpha1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha1"
 	clusterapiv1alpha2 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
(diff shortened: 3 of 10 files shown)


config/default/kustomization.yaml
```diff
@@ -1,12 +1,12 @@
 # Adds namespace to all resources.
-namespace: cluster-api-bootstrap-provider-kubeadm-system
+namespace: cabpk-system
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: cluster-api-bootstrap-provider-kubeadm-
```
chuckha: This prefix causes names > 63 characters to be created.

Makefile
```diff
@@ -1,6 +1,13 @@
 
-# Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+REGISTRY ?= gcr.io/$(shell gcloud config get-value project)
```
chuckha: This adds some niceness around testing

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -56,3 +87,31 @@ func (r *KubeadmBootstrapConfigReconciler) SetupWithManager(mgr ctrl.Manager) er
 		For(&kubeadmv1alpha1.KubeadmBootstrapConfig{}).
 		Complete(r)
 }
+func (r *KubeadmBootstrapConfigReconciler) Map(object handler.MapObject) []reconcile.Request {
```
rudoi: Can we use`ctrl.Request` here rather than `reconcile.Request`? Would save an import methinks.

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -56,3 +87,31 @@ func (r *KubeadmBootstrapConfigReconciler) SetupWithManager(mgr ctrl.Manager) er
...
+		return result
+	}
+
+	// There is no config reference for this machine, so don't enqueue anything
+	if m.Spec.Bootstrap.ConfigRef == nil {
+		return result
+	}
+
+	// Enqueue the associated config ref (not sure if this is actually what this should do?)
+	if m.Spec.Bootstrap.ConfigRef.Name != "" && m.Spec.Bootstrap.ConfigRef.Namespace != "" {
```
vincepri: Should we check if the APIGroup and Kind actually matches the object we're reconciling?

controllers/suite_test.go
```diff
@@ -62,6 +64,8 @@ var _ = BeforeSuite(func(done Done) {
 
 	err = kubeadmv1alpha1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
+	err = v1alpha2.AddToScheme(scheme.Scheme)
```
vincepri: Nit: This could be one line

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -36,17 +39,45 @@ type KubeadmBootstrapConfigReconciler struct {
...
 
+	cluster := &clusterapiv1alpha2.Cluster{}
+	clusterKey := types.NamespacedName{
+		Name:      machine.ClusterName,
+		Namespace: config.Namespace, // must still be in the same namespace
+	}
+	if err := r.Get(ctx, clusterKey, cluster); err != nil {
+		log.Error(err, "stacktrace", fmt.Sprintf("%+v", err))
+	}
+	log.Info("debug", "cluster", cluster, "machine", machine, "config", config)
```
vincepri: Extra?

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -36,17 +39,45 @@ type KubeadmBootstrapConfigReconciler struct {
...
+	}
+
 	machine := &clusterapiv1alpha2.Machine{}
-	if err := r.Get(ctx, req.NamespacedName, machine); err != nil {
+	machineKey := types.NamespacedName{
+		Name:      machineRef.Name,
+		Namespace: config.Namespace, // must be in the same namespace
+	}
+	if err := r.Get(ctx, machineKey, machine); err != nil {
 		log.Error(err, "stacktrace", fmt.Sprintf("%+v", err))
```
vincepri: Shouldn't we return an error here?

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -36,17 +39,45 @@ type KubeadmBootstrapConfigReconciler struct {
...
 	}
 
+	cluster := &clusterapiv1alpha2.Cluster{}
+	clusterKey := types.NamespacedName{
+		Name:      machine.ClusterName,
+		Namespace: config.Namespace, // must still be in the same namespace
+	}
+	if err := r.Get(ctx, clusterKey, cluster); err != nil {
+		log.Error(err, "stacktrace", fmt.Sprintf("%+v", err))
```
vincepri: Same here

controllers/kubeadmbootstrapconfig_controller.go
vincepri: Example: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/896/files#diff-fcb771956a4642d026c25a8ef9815e0bR155

chuckha: Yes, I'm still working on this code so I'll add it in a following PR that will actually do something interesting.

controllers/kubeadmbootstrapconfig_controller.go
chuckha: yeah, not necessary, will not show up in the following PRs

chuckha:
@vincepri appreciate the comments, I'm still working on this code which I didn't make obvious in the PR description so my fault. I should not have included the controller code at all but wanted to get the imports in and make sure go.mod was happy. The issues will be fixed when the controller actually does something

controllers/kubeadmbootstrapconfig_controller.go
chuckha: I'll do this in a following PR, I'm not 100% convinced any of this code is right. This PR should have been just the infrastructure changes/make/yaml changes. I tossed this code in just to get go.mod right, but  probably shouldn't have.

PR #26, opened 2019-07-15
Fix json tags on kubeadm types

chuckha:

This PR fixes the json tags in the bootstrap token. They should be ignored by json.Marshal and use the custom Unmarshal/Marshal functions instead.

Please see https://github.com/kubernetes/kubernetes/pull/80050

**Release note**:
```release-note
NONE
```


vincepri:
Should we also disallow to set the top `token` field given that you can't set any subfields of the object?

chuckha:
I don't think so, these objects still get populated due to the custom serializers

vincepri:
Doesn't the `json:"-"` tag make it so that the object never gets serialized or de-serialized?

chuckha:
No, it just ignores the default unmarshal/marshal functions. If the type fulfills the interface that still gets called.

https://play.golang.org/p/GFtkQkXN6-O

vincepri:
Very well then!

/lgtm

PR #28, opened 2019-07-15
Adds some tests and basic controller logic

chuckha:

We need this PR to get us started on writing more controller logic. This is super basic but it does add patterns for testing. We can *DEFINITELY* improve them and reduce boilerplate etc, but I'm curious about how people feel regarding the tests. Do they look like something you'd like to write and maintain? They are explicitly designed to unit test the controller. 

All code in this PR is fair game.

```release-note
NONE
```


controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -42,11 +44,28 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
+	// Find the owner reference
+	machineKind := v1alpha2.SchemeGroupVersion.WithKind("Machine").String()
+	var machineRef v1.OwnerReference
+	for _, ref := range config.OwnerReferences {
+		if ref.Kind == machineKind {
+			machineRef = ref
+			break
+		}
+	}
+	if machineRef.Kind == "" {
```
neolit123: non-blocking: alternatively machineRef could have been a pointer.

controllers/kubeadmbootstrapconfig_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+type log struct{}
+
+func (l *log) Info(msg string, keysAndValues ...interface{})             {}
+func (l *log) Enabled() bool                                             { return false }
+func (l *log) Error(err error, msg string, keysAndValues ...interface{}) {}
+func (l *log) V(level int) logr.InfoLogger                               { return l }
+func (l *log) WithValues(keysAndValues ...interface{}) logr.Logger       { return l }
+func (l *log) WithName(name string) logr.Logger                          { return l }
```
neolit123: do all unit tests in different files have to define a mock logger like so?

controllers/kubeadmbootstrapconfig_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	request := ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: "ns",
+			Name:      "cfg",
+		},
+	}
+	result, err := k.Reconcile(request)
+	if err != nil {
+		t.Fatal(fmt.Sprintf("%+v", err))
```
neolit123: do this error need a prefix e.g.:
`t.Fatalf("some message: %+v", err)`
[1]

controllers/kubeadmbootstrapconfig_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
```
neolit123: [1]

controllers/suite_test.go
```diff
@@ -1,79 +0,0 @@
...
-
-func TestAPIs(t *testing.T) {
-	RegisterFailHandler(Fail)
-
-	RunSpecsWithDefaultAndCustomReporters(t,
-		"Controller Suite",
-		[]Reporter{envtest.NewlineReporter{}})
-}
-
-var _ = BeforeSuite(func(done Done) {
```
vincepri: Any reason we prefer to not use the Kubebuilder reconcile suite?

controllers/kubeadmbootstrapconfig_controller_test.go
chuckha: Sometimes nil will suffice.

That being said, there is nothing stopping us from having a `bootstraptest` package that contains shared mocks _if the mocks are sufficiently dumb/reusable_. We do this in cluster-api-provider-aws with *generated* mocks, but shared mocks do not have to be generated.

Keeping mocks as thin and dumb as possible and not trying to reuse the same one for many kinds of functionality is crucial to maintaining clarity. Ideally mocks should contain no logic, but in practice they will need a tiny bit of logic.

controllers/kubeadmbootstrapconfig_controller_test.go
chuckha: It doesn't strictly need one and since the err is printing a stack trace of where the error comes from I didn't feel it was that important, but there's no harm in adding one. I'll do that.

controllers/kubeadmbootstrapconfig_controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	kubeadmv1alpha1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha1"
+	"sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type log struct{}
```
liztio: I'd prefer to see the logs only get spit out when the test fails, so failures are easier to debug

controllers/suite_test.go
chuckha: Do you mean `we` as in the community? I'm not sure which `we` you are referring to here.

 As commented in the PR description:

> The suite (that was deleted but can be undeleted) looks to do more integration level testing, but it didn't have any tests and we don't need any yet so I got rid of it.

I can add this back in if you think there's value here though

controllers/kubeadmbootstrapconfig_controller.go
chuckha: oh and then checked for nil? That's a good idea!

controllers/kubeadmbootstrapconfig_controller_test.go
chuckha: that's reasonable, this nil logger could print to stdout instead of doing nothing

That way the log lines will only be displayed on either a test failure or a verbose test run.

chuckha: ....or just use a real logger because we aren't testing log lines yet anyway

controllers/suite_test.go
chuckha: added back!

chuckha:
@vincepri  what is the hold for?

vincepri:
Feel free to remove it, I thought we were waiting on @fabriziopandini review

chuckha:
ah I did not mean to push those recent updates. those will go in a separate PR. One second.

Issue #16, opened 2019-07-09
Implement the controller logic for cloud-init output


chuckha:
/kind feature

**Describe the solution you'd like**
The output of this can be deduced from this paragraph in the cluster api proposal:

> Upon receiving a Machine event, a reconcile request is enqueued for the associated KubeadmBootstrapConfig. Next, it retrieves the Machine, Cluster, and generates a cloud-init based bootstrap configuration which is then stored in KubeadmBootstrapConfig.Status.BootstrapData. It also sets KubeadmBootstrapConfig.Status.Phase to “Ready”. A non-nil value in KubeadmBootstrapConfig.Status.BootstrapData is indicative to the Cluster API Machine Controller that the bootstrap provider has completed its work.

This means cloud-init is the first output. This does NOT mean cloud-init is the only output but it's a good way for us to get started.

**Anything else you would like to add:**

amy:
Lets split this up in 2 parts:

- [ ] Controller Logic
- [ ] Generating cloud-init configuration

We can use this issue to just track the progress of the controller logic part.

PR #30, opened 2019-07-16
Controller logic2

chuckha:
* This PR adds a few more steps in the controller logic.
* It improves the existing unit test. 
* It adds an integration test that exercises the code more thoroughly than a unit test can.
* It removes the Enum validation as that's not necessary right now
* Adds a script to copy in crds from CAPI

/assign @fabriziopandini 
/cc @amy

Part of #16 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/crd/bases/kubeadm.bootstrap.cluster.sigs.k8s.io_kubeadmbootstrapconfigs.yaml
+++ b/config/crd/bases/kubeadm.bootstrap.cluster.sigs.k8s.io_kubeadmbootstrapconfigs.yaml
@@ -1060,9 +1060,6 @@ spec:
               type: string
             phase:
               description: Phase is the state of the KubeadmBootstrapConfig object
-              enum:
-              - Ready
-              - Unknown
               type: string
           required:
           - phase
--- a/controllers/kubeadmbootstrapconfig_controller.go
+++ b/controllers/kubeadmbootstrapconfig_controller.go
@@ -29,6 +29,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var (
+	machineKind = v1alpha2.SchemeGroupVersion.WithKind("Machine").String()
+)
+
 // KubeadmBootstrapConfigReconciler reconciles a KubeadmBootstrapConfig object
 type KubeadmBootstrapConfigReconciler struct {
 	client.Client
@@ -46,12 +50,11 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 
 	config := kubeadmv1alpha1.KubeadmBootstrapConfig{}
 	if err := r.Get(ctx, req.NamespacedName, &config); err != nil {
-		log.Error(err, "stacktrace", fmt.Sprintf("%+v", err))
-		return ctrl.Result{}, errors.WithStack(err)
+		log.Error(err, "failed to get config", "stacktrace", fmt.Sprintf("%+v", err))
+		return ctrl.Result{}, err
 	}
 
 	// Find the owner reference
-	machineKind := v1alpha2.SchemeGroupVersion.WithKind("Machine").String()
 	var machineRef *v1.OwnerReference
 	for _, ref := range config.OwnerReferences {
 		if ref.Kind == machineKind {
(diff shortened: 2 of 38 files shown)


capi-config/BUILD.bazel
```diff
@@ -0,0 +1,22 @@
+filegroup(
```
chuckha: Please ignore all `capi-config/*` files! They are copied in from go mod and there is no filter (yet) so we get everything.

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -65,7 +68,31 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 	}
 
 	// Get the machine
+	machine := &v1alpha2.Machine{}
+	machineKey := client.ObjectKey{
+		Namespace: req.Namespace,
+		Name:      machineRef.Name,
+	}
+
+	if err := r.Get(ctx, machineKey, machine); err != nil {
```
chuckha: I will update the cluster-api dependency and use the util @vincepri pointed out that does this in a followup PR

controllers/kubeadmbootstrapconfig_controller_test.go
```diff
@@ -71,21 +71,32 @@ func (c *myClient) Status() client.StatusWriter {
...
-		db: map[string]runtime.Object{
-			"ns/cfg": &kubeadmv1alpha1.KubeadmBootstrapConfig{
-				ObjectMeta: metav1.ObjectMeta{
-					Namespace: "ns",
-					Name:      "cfg",
-					OwnerReferences: []metav1.OwnerReference{
-						{
-							Kind: machineKind,
-						},
+	objects := map[string]runtime.Object{
```
chuckha: The two tests look really similar right now, but the unit test gives us full control over the client. We can return errors and such that we cannot force through the k8s client. We also don't have to work schemes and such using the unit test. More control, but exercises fewer pieces.

config/crd/bases/kubeadm.bootstrap.cluster.sigs.k8s.io_kubeadmbootstrapconfigs.yaml
```diff
@@ -1060,9 +1060,6 @@ spec:
               type: string
             phase:
               description: Phase is the state of the KubeadmBootstrapConfig object
-              enum:
-              - Ready
-              - Unknown
               type: string
```
vincepri: Not sure if you need `Phase` here, IIRC we decided against Enums for the Phase field. `Status.Ready` is a required field though

chuckha: I thought the field was Phase and the value was Ready -- did that change to field is Ready and value is... true?

vincepri: Yeah, see https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/896/files#diff-e2e7d95efae5325d16854f456d43bcc3R85

controllers/kubeadmbootstrapconfig_controller.go
vincepri: This just merged https://github.com/kubernetes-sigs/cluster-api/pull/1159

config/crd/bases/kubeadm.bootstrap.cluster.sigs.k8s.io_kubeadmbootstrapconfigs.yaml
chuckha: ok, i see, where did that decision take place? I may be not watching a space I should be watching

controllers/suite_test.go
```diff
@@ -55,18 +53,18 @@ var _ = BeforeSuite(func(done Done) {
 
 	By("bootstrapping test environment")
 	testEnv = &envtest.Environment{
-		CRDDirectoryPaths: []string{filepath.Join("..", "config", "crd", "bases")},
+		CRDDirectoryPaths: []string{
+			filepath.Join("..", "config", "crd", "bases"),
+			filepath.Join("..", "capi-config", "crds"),
```
vincepri: You could use the vendored directory, but I see no vendor in this repo, this might need to get documented

config/crd/bases/kubeadm.bootstrap.cluster.sigs.k8s.io_kubeadmbootstrapconfigs.yaml
vincepri: I'll need to dig up some old CAPI PR, It was a while ago when we discussed with @ncdc @detiber @davidewatson and others about not using Phase, but instead switch to a single boolean field w.r.t. readiness. We've done the same upstream with `Status.InfrastructureReady` and `Status.BootstrapReady`.

This particular issue is a documentation one that should be addressed once we figure out the CAPI gitbook story

chuckha: i'm surprised the proposal wasn't updated -- is that pretty much out of date? will I have to dig through all the old CAPI PRs to find more decisions like this?

vincepri: I can open a PR to update the proposal, afaik other changes have been synced

hack/update-capi-crds.sh
```diff
@@ -0,0 +1,31 @@
...
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# find the module version and the location on the system
+src=$(go mod download -json sigs.k8s.io/cluster-api  | jq -r .Dir)
+
+# -p will not fail if the dir exists
+mkdir -p capi-config
```
fabriziopandini: what about using `config-capi` instead so we can get all the folders containing config files close together  ...

chuckha: yes! that is a great idea.

chuckha:
ok @fabriziopandini this is ready to go!

PR #31, opened 2019-07-16
Implement main.go

chuckha:
* Remove dependency on kustomize (use kubectl kustomize instead)

This PR implements main straight out of Kubebuilder book except for our types.

This removes the dependency on kustomize when installing the manager and instead uses `kubectl kustomize`.

**Release note**:
```release-note
NONE
```

/assign @fabriziopandini 
/cc @amy @neolit123 @vincepri

### DIFF
--- a/main.go
+++ b/main.go
@@ -19,9 +19,55 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
+	"flag"
+	"os"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"k8s.io/klog/klogr"
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha1"
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/controllers"
+	ctrl "sigs.k8s.io/controller-runtime"
+	// +kubebuilder:scaffold:imports
 )
 
+var (
+	myscheme = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+func init() {
+	_ = scheme.AddToScheme(myscheme)
+	_ = v1alpha1.AddToScheme(myscheme)
+	// +kubebuilder:scaffold:scheme
+}
+
 func main() {
-	fmt.Println("hello world")
+	var metricsAddr string
+	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.Parse()
+
+	ctrl.SetLogger(klogr.New())
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{Scheme: myscheme, MetricsBindAddress: metricsAddr})
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	if err := (&controllers.KubeadmBootstrapConfigReconciler{
+		Client: mgr.GetClient(),
+		Log:    ctrl.Log.WithName("reconciler"),
+	}).SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "reconciler")
+		os.Exit(1)
+	}
+	// +kubebuilder:scaffold:builder
+
+	setupLog.Info("starting manager")
+	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+		setupLog.Error(err, "problem running manager")
+		os.Exit(1)
+	}
 }
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ install: manifests
 # Deploy controller in the configured Kubernetes cluster in ~/.kube/config
 deploy: manifests
 	kubectl apply -f config/crd/bases
-	kustomize build config/default | kubectl apply -f -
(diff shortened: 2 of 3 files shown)


chuckha:
I built, pushed and ran it on a cluster but did not exercise the watch at all.

main.go
```diff
@@ -19,9 +19,56 @@ limitations under the License.
...
 )
 
+var (
+	myscheme = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
+)
+
+func init() {
+	_ = scheme.AddToScheme(myscheme)
+	_ = v1alpha1.AddToScheme(myscheme)
```
neolit123: is this something that kubebuilder requires?
would it matter if these calls are under `var (` or `main()` with the idea to not have an `init()` function in this file?

chuckha: this was mostly a straight port from the tube builder book. Doesn't matter where it is to be honest. I think this is for the `// +kubebuilder:scaffold:scheme` thing, which I don't actually know what that is...Let me look into that and see if we need this to be its own section or what.

chuckha: Ah yes, here it is: https://github.com/kubernetes-sigs/kubebuilder/blob/master/pkg/scaffold/v2/main.go#L66

This section doesn't have to be in init() but if we end up generating more types the scaffold will try and put stuff here. Might as well keep it how they intended it to be imo

neolit123: ok, understood. having `init()` in main.go in general seems redundant, but if the tooling uses it, then it's fine.

main.go
```diff
@@ -19,9 +19,56 @@ limitations under the License.
...
+
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/controllers"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/scheme"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	// +kubebuilder:scaffold:imports
```
vincepri: What's this for?

main.go
```diff
@@ -19,9 +19,56 @@ limitations under the License.
...
+	// +kubebuilder:scaffold:scheme
+}
+
 func main() {
-	fmt.Println("hello world")
+	var metricsAddr string
+	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.Parse()
+
+	ctrl.SetLogger(log.ZapLogger(true))
```
vincepri: klogr?

main.go
```diff
@@ -19,9 +19,56 @@ limitations under the License.
...
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{Scheme: myscheme, MetricsBindAddress: metricsAddr})
+	if err != nil {
+		setupLog.Error(err, "unable to start manager")
+		os.Exit(1)
+	}
+
+	if err = (&controllers.KubeadmBootstrapConfigReconciler{
```
vincepri: `if err :=`

main.go
```diff
@@ -19,9 +19,56 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
+	"flag"
+	"os"
+
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/controllers"
+
```
vincepri: Extra space

main.go
chuckha: kubebuilder uses these scaffold markers as a hook to add more things during a scaffold create action (such as making a new crd)

main.go
chuckha: klogr requires flag parsing and setting v and is overall a bad experience. We already have this exported with controller-runtime. What benefits do you see with klogr?

main.go
chuckha: grrrr i need to figure out what the HECK goland/goimports thinks it's doing

main.go
chuckha: yes, thx

main.go
chuckha: I think @ncdc had suggested zap at some point in the recent past? Maybe i'm mis-remembering though.

anyway, not a huge deal. If you really want me to change it I will, i'm like a 5/10 on the caring-about-it scale.

main.go
vincepri: Interesting thanks!

main.go
vincepri: I wish we could use similar tools and patterns across all CAP* repos, I'm fine with either, I would just like to use one

ncdc: kubebuilder v2 uses zap by default in its scaffolding: https://github.com/kubernetes-sigs/kubebuilder/blob/fb5789a12a651e07172e3d509d71d117e37c29cd/pkg/scaffold/v2/main.go#L166

+1 to the same logger in the ecosystem. Maybe we can help everyone move to kubebuilder v2 & zap.

chuckha:
/hold cancel

this is ok for another review if anyone has any final comments

main.go
neolit123: the general agreement on the sig-arch mailing list (some time ago; don't have the links) was to use klog / klogr for everything.

chuckha: For this PR I'm going to stick with zap. if we need to change it for some reason later we can.

chuckha:
Updated to use klogr instead of zap. I don't think we're ready to have the zap fight yet.

PR #33, opened 2019-07-17
Move control plane locker from CAPA

### LINKED ISSUE #29: Move control plane locker from CAPA
/kind feature

The [ControlPlaneInitLocker](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/release-0.3/pkg/cloud/aws/actuators/machine/control_plane_init_locker.go) has been merged into CAPA to prevent races to spin up multiple control plane nodes in the same cluster, we should port this functionality to the kubeadm bootstrap controller logic to make sure we only create a single control plane and prevent any race conditions.

SataQiu:
Move control plane locker from CAPA

Fixes #29

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/control_plane_init_locker.go
+++ b/controllers/control_plane_init_locker.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	apicorev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	clusterv2 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
+)
+
+// ControlPlaneInitLocker provides a locking mechanism for cluster initialization.
+type ControlPlaneInitLocker interface {
+	// Acquire returns true if it acquires the lock for the cluster.
+	Acquire(cluster *clusterv2.Cluster) bool
+}
+
+// controlPlaneInitLocker uses a ConfigMap to synchronize cluster initialization.
+type controlPlaneInitLocker struct {
+	log             logr.Logger
+	configMapClient corev1.ConfigMapsGetter
+}
+
+var _ ControlPlaneInitLocker = &controlPlaneInitLocker{}
+
+func newControlPlaneInitLocker(log logr.Logger, configMapClient corev1.ConfigMapsGetter) *controlPlaneInitLocker {
+	return &controlPlaneInitLocker{
+		log:             log,
+		configMapClient: configMapClient,
+	}
+}
+
+func (l *controlPlaneInitLocker) Acquire(cluster *clusterv2.Cluster) bool {
+	configMapName := fmt.Sprintf("%s-controlplane", cluster.UID)
+	log := l.log.WithValues("namespace", cluster.Namespace, "cluster-name", cluster.Name, "configmap-name", configMapName)
+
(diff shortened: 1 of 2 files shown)


chuckha:
Hmm this linter is actually failing. I need to make an issue to look into that. In the meantime can you run `./hack/verify-all.sh` locally and fix the issues?

SataQiu:
OK, I will do that.  Thanks @chuckha

chuckha:
Ping me in a comment when this is passing and ready for review! sorry about the bad signal with CI, I don't know what's going on there.

chuckha:
/hold

This is failing the verify scripts.

ashish-amarnath:
This needs to be fixed. Automated tests are not catching this atm 
```
Running golint...
cloudinit/cloudinit.go:32:6: exported type BaseUserData should have comment or be unexported
Found 1 lint suggestions; failing.
cloudinit/controlplane_certs.go:21:6: exported type Certificates should have comment or be unexported
Found 1 lint suggestions; failing.
```

controllers/control_plane_init_locker.go
```diff
@@ -0,0 +1,104 @@
...
+var _ ControlPlaneInitLocker = &controlPlaneInitLocker{}
+
+func newControlPlaneInitLocker(log logr.Logger, configMapClient corev1.ConfigMapsGetter) *controlPlaneInitLocker {
+	return &controlPlaneInitLocker{
+		log:             log,
+		configMapClient: configMapClient,
+	}
+}
+
+func (l *controlPlaneInitLocker) Acquire(cluster *clusterv2.Cluster) bool {
```
fabriziopandini: Might be I'm missing something, but I don't see the counterpart of Acquire (Release)

vincepri: In the original codebase, the release part was done in the Cluster Actuator https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/actuators/cluster/actuator.go#L140-L159, we should move this as well

vincepri:
@fabriziopandini This codebase originally solved the problem of generating multiple Kubeadm control-plane init configurations when a user would create an HA cluster submitting multiple Machine instances at the same time. 

I see how it may touch a little outside of its primary responsibility: generating cloud-init data. That said, without coordination, we might end up generating incorrect data. Happy to chat more in Zoom or on Slack if you want to talk more about it.

/cc @ncdc

ncdc:
Longer term, when we tackle control plane management (v1alpha3, perhaps), I'd expect this code to go away and be replaced by a new component that manages the full lifecycle of all control plane machines (init, scale up (join), scale down, upgrade, etc).

ncdc:
Also, we need to orchestrate when the infrastructure provider is allowed to create a server:

1. If there is no control plane yet, we should only generate the bootstrap data for a machine if the machine is a control plane machine and we are able to acquire the control plane init lock
1. If there is a control plane, we can generate the bootstrap data immediately

controllers/control_plane_init_locker.go
SataQiu: `Reconcile` is not complete yet. I think we can add this later. Right?

fabriziopandini:
@vincepri @ncdc thanks for the clarification! ok for me

controllers/control_plane_init_locker.go
fabriziopandini: @SataQiu given that the Release code pointed out by @vincepri is well-scoped, IMO it would be great to include it in this PR so the control plane locker is consistent.
wdyt?

SataQiu:
I'm confused. Why did `pull-cluster-api-bootstrap-provider-kubeadm-verify` fail ？
Did I miss something? @chuckha 
```sh
[*] Verifying whitespace...
[*] Verifying spelling...
[*] Verifying boilerplate...
[*] Verifying gofmt...
[*] Verifying golint...
[*] Verifying govet...
[*] Verifying deps...
[*] Verifying gotest...
[*] Verifying build...
All verify checks passed, congrats! 
```

controllers/control_plane_init_locker.go
```diff
@@ -0,0 +1,126 @@
...
+
+		// Unable to acquire
+		return false
+	}
+
+	// Successfully acquired
+	return true
+}
+
+func (l *controlPlaneInitLocker) Release(cluster *clusterv2.Cluster) bool {
```
chuckha: interesting, golint doesn't fail this, presumably because controlPlaneInitLocker is not exported.

controllers/control_plane_init_locker.go
chuckha: I'll move it over when I need it in the actual usage of this code. no worries here.

chuckha:
/hold cancel
/approve

whoops, ignore the lgtm, this already had lgtm

chuckha:
Thank you @SataQiu for your first PR into CABPK! Welcome aboard :) 🚂 🚀

PR #34, opened 2019-07-17
Make KubeadmBootstrapConfig type to use kubeadm config v1beta2

fabriziopandini:
CAPPK `KubeadmBootstrapConfig` type shoud use kubeadm config version `v1beta2` instead of `v1beta1`.

`v1beta2` is the latest version and it contains few improvements including better usage of the "omitempty" tag. This can improve significatively the experience of users writing yaml files with the `KubeadmBootstrapConfig` objects.

In fact, with v1beta2 the minimal `KubeadmBootstrapConfig ` object that passes kubectl validation is 
```
apiVersion: kubeadm.bootstrap.cluster.sigs.k8s.io/v1alpha1
kind: KubeadmBootstrapConfig
metadata:
  name: kubeadmbootstrapconfig-sample
spec:
  clusterConfiguration:
    apiVersion: kubeadm.k8s.io/v1beta2
    kind: ClusterConfiguration
```

while with v1beta1 the user experience is much much worse
```
apiVersion: kubeadm.bootstrap.cluster.sigs.k8s.io/v1alpha1
kind: KubeadmBootstrapConfig
metadata:
  name: kubeadmbootstrapconfig-sample
spec:
  clusterConfiguration:
    apiVersion: kubeadm.k8s.io/v1beta2
    kind: ClusterConfiguration
    etcd:   
      local:
        dataDir: ""  
        extraArgs:
          a: "a"
    networking:
      podSubnet: ""
      serviceSubnet: ""
      dnsDomain: ""
    kubernetesVersion: ""
    controlPlaneEndpoint: ""
    dns:
      type: "CoreDNS"
    certificatesDir: ""
    imageRepository: ""
```

### DIFF
--- a/api/v1alpha1/kubeadmbootstrapconfig_types.go
+++ b/api/v1alpha1/kubeadmbootstrapconfig_types.go
@@ -18,7 +18,7 @@ package v1alpha1
 
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kubeadmv1beta1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
+	kubeadmv1beta2 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta2"
 )
 
 // Phase defines KubeadmBootstrapConfig phases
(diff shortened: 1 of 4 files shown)


fabriziopandini:
It seems that `envtest.Environment` handles validation in a different way that the real api server. In the unit test, and I was forced to assign a value to `nodeRegistration.taints` both in `InitConfiguration` and `JoinConfiguration`
:thinking

chuckha:
I was a little overoptimistic on this one. I believe, because of our AMIs/machine images, we pin kubeadm and kubelet to the same version. In order to install a 1.14 we'd need the kubelet to be 1.14 and that won't happen if kubeadm is 1.15.

Going to close this for now but I really want us to use v1beta2 asap.

/close

Issue #27, opened 2019-07-15
implement a machine watch to trigger a config reconcile


chuckha:
/kind feature

**Describe the solution you'd like**
As part of #16 we will have to watch machine objects and trigger the associated config object to get reconciled. This is using kubebuilder's MapFn feature.

/assign @amy
/lifecycle active

**Anything else you would like to add:**
We do something similar in CAPI here: https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/controller/machineset/machineset_controller.go#L110

That idea will need to be ported to kubebuilder v0.2.

The goal is that if you run the CABPK controller and create a machine with an associated kubeadm bootstrap config you will see that we reconcile the associated bootstrap config on a machine update.

vincepri:
It'd be great to add this upstream, like https://github.com/kubernetes-sigs/cluster-api/pull/1153. It has the benefit that the mapping functions can be shared between all providers

chuckha:
That sounds reasonable to me! thanks for the PR link.

chuckha:
/close

please see #39 for details

PR #39, opened 2019-07-17
Add a watch for machines to config mapping

amy:
Adds a watch queue for new machine objects to trigger actions on the config objects.

Addresses #27

### DIFF
--- a/controllers/kubeadmbootstrapconfig_controller.go
+++ b/controllers/kubeadmbootstrapconfig_controller.go
@@ -23,10 +23,13 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	kubeadmv1alpha1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha1"
 	"sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 var (
@@ -55,6 +58,7 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 	}
 
 	// Find the owner reference
+	machineKind := v1alpha2.SchemeGroupVersion.WithKind("Machine").String()
 	var machineRef *v1.OwnerReference
 	for _, ref := range config.OwnerReferences {
 		if ref.Kind == machineKind {
@@ -100,5 +104,12 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 func (r *KubeadmBootstrapConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&kubeadmv1alpha1.KubeadmBootstrapConfig{}).
+		Watches(&source.Kind{Type: &v1alpha2.Machine{}}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: MachineToConfigMapFunc(schema.GroupVersionKind{
+				Group:   v1alpha2.SchemeGroupVersion.Group,
+				Version: v1alpha2.SchemeGroupVersion.Version,
+				Kind:    "Machine",
+			}),
+		}).
 		Complete(r)
 }
--- a/controllers/watch.go
+++ b/controllers/watch.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	capiv1alpha2 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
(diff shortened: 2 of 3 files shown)


amy:
This is a WIP. Will tag people after I've added tests and gone through the code some more.

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -23,12 +23,61 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+// Add creates a new KubeadmBootstrap Controller and adds it to the Manager with default RBAC. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager) error {
```
chuckha: We don't need any of this, this is all replaced in controller-runtime v0.2. Instead of this use the .For function on the manager builder

controllers/watch.go
```diff
@@ -0,0 +1,51 @@
...
+// Machine events and returns reconciliation requests for a Configuration object.
+func MachineToConfigMapFunc(gvk schema.GroupVersionKind) handler.ToRequestsFunc {
+	return func(o handler.MapObject) []reconcile.Request {
+		m, ok := o.Object.(*capiv1alpha2.Machine)
+		if !ok {
+			return nil
+		}
+
+		// Return early if the GroupVersionKind doesn't match what we expect.
+		infraGVK := m.Spec.InfrastructureRef.GroupVersionKind()
```
vincepri: This should reference Bootstrap.ConfigRef instead of InfrastructureRef, it'd be great to contribute this upstream though in the util package, so others can use it as well

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -23,12 +23,61 @@ import (
...
+		&source.Kind{Type: &capiv1alpha2.Machine{}},
+		&handler.EnqueueRequestForObject{},
+	)
+	if err != nil {
+		return err
+	}
+
+	return c.Watch(
+		&source.Kind{Type: &capiv1alpha2.Machine{}},
+		&handler.EnqueueRequestsFromMapFunc{
```
vincepri: This should be moved to in SetupWithManager using https://github.com/kubernetes-sigs/controller-runtime/blob/bace65823215cf3226bd0395fab06f0fe798c935/pkg/builder/controller.go#L99

controllers/watch.go
amy: Good catch! Re: configref.
Gonna get this merged first. Then make an issue to move the function upstream.

vincepri: You might need to check that the ConfigRef is not-nil before calling `GroupVersionKind()`, because, in case of bootstrap providers, it's optional.

chuckha:
looks fine to me. i think we're nearly at a point of testing when the reconcile gets called

this looks like it needs a rebase before it can merge though

vincepri:
Seems my comment was missed https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/39#discussion_r304647393

amy:
> Seems my comment was missed [#39 (comment)](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/39#discussion_r304647393)

Nope. Wasn't missed. Just working on tests! In the future, please start a new comment instead of replying to a resolved comment. Will update in the next PR thanks for the review.

controllers/watch.go
```diff
@@ -0,0 +1,67 @@
...
+		m, ok := o.Object.(*capiv1alpha2.Machine)
+		if !ok {
+			return nil
+		}
+
+		// ConfigRef is an optional field
+		configRef := m.Spec.Bootstrap.ConfigRef
+
+		if configRef == nil {
+			return []reconcile.Request{
```
chuckha: i'd expect this to return nil, why is it returning a request to reconcile for a blank name?

controllers/watch_test.go
```diff
@@ -0,0 +1,139 @@
...
+		output  []reconcile.Request
+	}{
+		{
+			name: "reconcile infra-1",
+			input: schema.GroupVersionKind{
+				Group:   "foo.cluster.sigs.k8s.io",
+				Version: "v1alpha2",
+				Kind:    "TestMachine",
+			},
+			request: handler.MapObject{
```
chuckha: It looks like the only thing changed between all of these test cases is the config ref, could we only supply the config ref in the table and build the rest of this inside the subtest to reduce some of this boilerplate that doesn't appear to change?

controllers/watch.go
amy: oh. Fixing!

controllers/watch_test.go
amy: yurp

chuckha:
/approve
/assign @ncdc 
for lgtm

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -100,5 +104,12 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 func (r *KubeadmBootstrapConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&kubeadmv1alpha1.KubeadmBootstrapConfig{}).
+		Watches(&source.Kind{Type: &v1alpha2.Machine{}}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: MachineToConfigMapFunc(schema.GroupVersionKind{
```
ncdc:
```suggestion
			ToRequests: MachineToConfigMapFunc(v1alpha2.SchemeGroupVersion.WithKind("Machine")),
```

controllers/watch_test.go
```diff
@@ -0,0 +1,108 @@
...
+	var testcases = []struct {
+		name    string
+		input   schema.GroupVersionKind
+		request *corev1.ObjectReference
+		output  []reconcile.Request
+	}{
+		{
+			name: "reconcile infra-1",
+			input: schema.GroupVersionKind{
+				Group:   "foo.cluster.sigs.k8s.io",
```
ncdc: Today, with the guidance from SIG Architecture, we decided to use `x-k8s.io` instead of `sigs.k8s.io`. Please make this change anywhere you have `sigs.k8s.io` in this PR. If there are other places in the code that use the old name, we can open a new PR to fix them. Thanks.

chuckha:
this is gonna need a hella rebase after #49 merges. our crd types changed

chuckha:
we don't need this because if the machine updates it will already have the bootstrap data so we will never need to re-reconcile config if we see a machine change

/close

Issue #18, opened 2019-07-10
Cloud-init script generation


amy:
This is split off from #16. We need a stand-alone or existing library that will generate cloud-init scripts for this provider.

chuckha:
We can consider moving the existing code from CAPA to here that does this already

moshloop:
I think while templates work, they are not ideal in the long term as evidenced by the kops migration away from templates to structs. 

There are golang structs for cloud-init that could be copied and pasted from here: https://github.com/moshloop/konfigadm/blob/master/pkg/cloud-init/types.go 

Another option is to import konfigadm as whole as it will enable other functionality such as embedding arbitrary files, multi-os  installation of prerequisites like  kubernetes/docker, coreos ignition support,

chuckha:
Agreed, templates are definitely not ideal in the long term. I think they are a fine place to start and as the project expands and we determine the scope more clearly we can talk about alternative methods such as konfigadm. That being said, this library (bootstrap-provider-kubeadm) should be built so that changing out the script generation library is not painful.

rudoi:
I think starting with what we have now in CAPA and refining makes the most sense to me, though I agree structs are the future. I'll take a swing at this. 😄 

/assign
/lifecycle active

PR #42, opened 2019-07-18
Generate kubeadm cloud init code

### LINKED ISSUE #16: Implement the controller logic for cloud-init output

chuckha:

This PR generates the control-plane-init cloud init script, but it is incomplete. I just ran out of time this week and wanted to get some code out for folks to look at and think about.

Related to #18 
Closes #16 

Definitely not done. Please see TODOs

**Release note**:
```release-note
NONE
```

/cc @amy @vincepri @liztio @ncdc @detiber


controllers/control_plane_init.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+type UserDataGenerator struct {
+	logr.Logger
+}
+
+func (u *UserDataGenerator) GenerateControlPlaneInit(config *v1alpha1.KubeadmBootstrapConfig) error {
+	u.Info("Generating user data for first control plane")
+
+	// TODO: sub in things in the cluster configuration like the cluster name and other data we can extract from the cluster object
```
chuckha: In this version of the code we don't support custom certificates. However, there is a bunch of defaulting I can do to make this config actually work. The user must still supply

* Controlplane endoint + port formatted correctly
* cloud provider

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -93,6 +98,30 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
+	if r.UserDataGenerator == nil {
+		r.UserDataGenerator = &UserDataGenerator{
+			Logger: r.Log.WithName("user-data-generator"),
+		}
+	}
+	if err := r.UserDataGenerator.GenerateControlPlaneInit(&config); err != nil {
+		log.Error(err, "error generating control plane init")
+		return ctrl.Result{}, err
+	}
+	// TODO save config object
```
chuckha: This TODO would make the code "work"

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -93,6 +98,30 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
 		log.Error(err, "failed to get cluster")
 		return ctrl.Result{}, errors.WithStack(err)
 	}
+
+	// If the machine is not Ready, requeue until it's ready
```
ncdc: This needs to be removed - the machine controller determines bootstrapReady based on the value in the bootstrap resource's `status.ready` field (which, in this case, is this code here!).

controllers/control_plane_init.go
```diff
@@ -0,0 +1,48 @@
...
+	config.Spec.ClusterConfiguration.APIVersion = "v1beta1"
```
fabriziopandini: What if the users pass a wrong value in this field? are we overriding silently or should we fail/warn

controllers/control_plane_init.go
```diff
@@ -0,0 +1,48 @@
...
+	config.Spec.ClusterConfiguration.Kind = "ClusterConfiguration"
+	clusterdata, err := json.Marshal(config.Spec.ClusterConfiguration)
```
fabriziopandini: I'm a little bit confused about this because `GenerateControlPlaneInit` in part is setting/enforcing values in the config, and in other parts it is consuming the config (like here)
What about moving all the consume config part into `cloudinit.NewInitControlPlane`?

controllers/control_plane_init.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"encoding/json"
+
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha1"
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/cloudinit"
+)
+
+type UserDataGenerator struct {
```
fabriziopandini: What about `BootstrapDataGenerator` or `CloudInitGenerator`?

controllers/kubeadmbootstrapconfig_controller.go
```diff
@@ -93,6 +98,30 @@ func (r *KubeadmBootstrapConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Res
...
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+	}
+
+	// if the bootstrap data is not nil then it's already been generated and we don't have anything else to do
+	if machine.Spec.Bootstrap.Data != nil {
+		log.Info("Nothing to be done for this machine", "machine-name", machine.Name)
+		return ctrl.Result{}, nil
+	}
+
+	// Generate the cloud-init for controlplane init
```
fabriziopandini: Am I wrong or we should also expect the UserDataGenerator handle also controlpane join and node join?

controllers/control_plane_init.go
```diff
@@ -0,0 +1,48 @@
+package controllers
+
+import (
+	"encoding/json"
```
fabriziopandini: sigs.k8s.io/yaml?

controllers/control_plane_init.go
chuckha: Yes, BootstrapDataGenerator is good!

controllers/control_plane_init.go
chuckha: In the past we have issued a message when a user tried to provide default value that did not match what our code enforces.

I believe we should do the same and issue a log

controllers/control_plane_init.go
chuckha: That's fine, I hadn't spent time structuring this code. I plan to improve this as we add more than just control-plane-init generation and if that refactor makes sense I will use it.

controllers/kubeadmbootstrapconfig_controller.go
chuckha: ahh ok, so once the bootstrap config becomes ready then the machine controller will update the machine.status.BootstrapReady to true. yes, this should be removed in that case.

controllers/kubeadmbootstrapconfig_controller.go
chuckha: you are 100% correct. 

As stated in the description:

> This PR generates the control-plane-init cloud init script, but it is incomplete.

fabriziopandini: oh sorry, I missed that

controllers/control_plane_init.go
chuckha: Maybe. Depending on if we need that much power. Right now it should be sufficient to generate the configurations as JSON docs, but we may end up needing yaml at some point.

ncdc:
@chuckha do you need any help with this?

fabriziopandini:
@cha if you need some help here, I'm happy to jump in

chuckha:
/assign @fabriziopandini 

This is ready to unblock you!

controllers/kubeadmconfig_controller.go
```diff
@@ -108,16 +122,29 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
-	config.Status.BootstrapData = []byte("hello world")
-	config.Status.Ready = true
+	// Generate the cloud-init for controlplane init
+	if r.BootstrapDataGenerator == nil {
+		r.BootstrapDataGenerator = &BootstrapDataGenerator{
+			Logger: r.Log.WithName("user-data-generator"),
+		}
+	}
+
+	if util.IsControlPlaneMachine(machine) {
```
ncdc: Do we want to do this, or just check to see if `config` has the `ClusterConfiguration` and `InitConfiguration` fields set? Or both?

controllers/kubeadmconfig_controller.go
```diff
@@ -108,16 +122,29 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+		locked := r.ControlPlaneInitLock.Acquire(cluster)
+		if locked {
```
ncdc: How about inverting the check to minimize nesting:
```go
if !locked {
  return ctrl.Result{RequeueAfter: someDuration}, nil
}
// Generate control plane init
// ...
```

locker/control_plane_init_locker.go
```diff
@@ -27,28 +27,15 @@ import (
...
-func newControlPlaneInitLocker(log logr.Logger, configMapClient corev1.ConfigMapsGetter) *controlPlaneInitLocker {
-	return &controlPlaneInitLocker{
-		log:             log,
-		configMapClient: configMapClient,
-	}
-}
-
-func (l *controlPlaneInitLocker) Acquire(cluster *clusterv2.Cluster) bool {
+// Acquire will attempt to acquire a lock that guarantees this is the first control plane to be created.
+// TODO Consider adding a short circuit if there is already a control plane node to reduce the calls to the API server.
```
ncdc: What kind of short circuit did you have in mind? This code does a "get" check first. Were you thinking of something else?

controllers/kubeadmconfig_controller.go
detiber: Switching to checking for ClusterConfiguration/InitConfiguration would potentially remove the need for a lock completely.

ncdc: We'd still need it. If you submit multiple control plane machines, or worker machines, everything except the init has to wait.

detiber: Yes, the other Machines need to wait, but that doesn't necessitate a lock, only that they wait for the "Control Plane Ready" annotation. The lock was only needed for determining which control plane Machine should do the init vs join.

ncdc: Fair point

locker/control_plane_init_locker.go
chuckha: like, if there is already a control plane node we don't need to worry about locks any more since control planes can be joined in parallel as well as machines. But that is very likely a premature optimization.

ncdc: Given the discussion in https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/42#discussion_r310661445, we may no longer need the locker code. TBD. WDYT?

controllers/kubeadmconfig_controller.go
chuckha: Hold up, lemme try to rephrase to check my understanding:

There is only one reconciler running at a time and it is single threaded so we'll reconcile machines one at a time. This makes locking pretty much already not necessary.

The first control plane machine that comes in gets the clusterconfiguration/initconfiguration and annotation is added to some object (?) "control-plane-ready: yes". Every following control-plane machine that gets reconciled checks this annotation and will be a control-plane join. Workers are always joins.

I'm trying to figure out what object to store this annotation on and if "control-plane-ready" is the right name for this. It doesn't feel right, unless perhaps the value becomes a state (no => no control plane has been created so the current machine can be the first; pending => a control plane has been assigned and is being init'd => yes the control plane is up and running)

ncdc: > There is only one reconciler running at a time and it is single threaded so we'll reconcile machines one at a time. This makes locking pretty much already not necessary.

We shouldn't assume this is always going to be the case.

Also, with CAPA & v1alpha1, it can finish reconciling a Machine and move on to the next Machine before the control plane is ready, so we need this in v1alpha1, even with it being single-threaded.

> The first control plane machine that comes in gets the clusterconfiguration/initconfiguration

This would be manually configured by the user.

> annotation is added to some object (?) "control-plane-ready: yes".

For CAPA v1alpha1, we put this on the `Cluster`.

> Every following control-plane machine that gets reconciled checks this annotation and will be a control-plane join. Workers are always joins.

The user would be expected to only fill in `KubeadmConfig.Spec.JoinConfiguration` for these.

> I'm trying to figure out what object to store this annotation on and if "control-plane-ready" is the right name for this. It doesn't feel right, unless perhaps the value becomes a state (no => no control plane has been created so the current machine can be the first; pending => a control plane has been assigned and is being init'd => yes the control plane is up and running)

If you want an alternative that doesn't require an annotation, we could list all the Machines for the Cluster, and if at least one of them has status.nodeRef set, that means the control plane is ready (the nodeRef is only set if we can reach the apiserver).

chuckha: Ahh of course the user sets it. yeah, that absolutely works for me.

controllers/bootstrapdata.go
```diff
@@ -0,0 +1,67 @@
...
+
+// GenerateControlPlaneInit will generate the bootstrap data and set the fields necessary on the config object.
+func (u *BootstrapDataGenerator) GenerateControlPlaneInit(config *v1alpha2.KubeadmConfig) error {
+	u.Info("Generating user data for first control plane")
+
+	// TODO: use secrets for certs, plug into kubeadm
+	// TODO: sub in things in the cluster configuration like the cluster name and other data we can extract from the cluster object
+	config.Spec.ClusterConfiguration.APIVersion = "v1beta1"
+	config.Spec.ClusterConfiguration.Kind = "ClusterConfiguration"
+	clusterdata, err := json.Marshal(config.Spec.ClusterConfiguration)
```
fabriziopandini: @chuckha 
There is a potential overlap between network settings in `CAPBK.v1alpha2.KubeadmConfig.ClusterConfiguration` and  network settings in `CAPI.v1alpha2.Cluster`. What is expected behavior here? (once I know the behavior, I can fix it also in a fallow PR)

fabriziopandini: Same goes e.g for the `CAPBK.controlplaneEndpoint`/ `CAPI.APIEndpoints`

chuckha: yes, you're right. I'll open a ticket to address this later

chuckha: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/78

fabriziopandini:
Thanks! Let's get this merged and move on with the next iteration
/lgtm

controllers/kubeadmconfig_controller.go
```diff
@@ -111,9 +122,87 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+		cloudInitData, err := cloudinit.NewInitControlPlane(&cloudinit.ControlPlaneInput{
+			InitConfiguration:    initdata,
+			ClusterConfiguration: clusterdata,
+		})
+		if err != nil {
+			log.Error(err, "failed to generate cloud init for bootstrap control plane")
+			return ctrl.Result{}, err
+		}
+
+		config.Status.BootstrapData = cloudInitData
```
detiber: I just noticed that BootstrapData is defined as []byte, does this have implications on serialization that we need to worry about?

PR #53, opened 2019-07-26
Enable leader election in manager

chuckha:

This PR allows the manager to work with with the default `--enable-leader-election` flag that kubebuilder has added to the Deployment YAML.

**Release note**:
```release-note
Enable highly available manager
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -45,12 +45,19 @@ func init() {
 
 func main() {
 	var metricsAddr string
+	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
+		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.Parse()
 
 	ctrl.SetLogger(klogr.New())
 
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -45,12 +45,19 @@ func init() {
...
+		"Enable leader election for controller manager. Enabling this will ensure there is only one active controller manager.")
 	flag.Parse()
 
 	ctrl.SetLogger(klogr.New())
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{Scheme: myscheme, MetricsBindAddress: metricsAddr})
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:             myscheme,
+		MetricsBindAddress: metricsAddr,
+		LeaderElection:     enableLeaderElection,
```
detiber: Should `LeaderElectionNamespace` also be set? https://github.com/kubernetes-sigs/controller-runtime/blob/59b131b7cd54d56ec74a66b084a53dd1c3e4843f/pkg/manager/manager.go#L113

chuckha: No, it should be detected automatically:

https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/leaderelection/leader_election.go#L61

PR #56, opened 2019-07-31
Ignore non-pending machines

chuckha:

This PR ignores machines that are not in Pending mode

/assign @ncdc @detiber 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -27,7 +27,7 @@ rules:
   - patch
   - update
 - apiGroups:
-  - cluster.sigs.k8s.io
+  - cluster.x-k8s.io
   resources:
   - clusters
   - machines
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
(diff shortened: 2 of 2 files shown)


controllers/kubeadmconfig_controller.go
```diff
@@ -79,6 +78,11 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 		return ctrl.Result{}, errors.WithStack(err)
 	}
 
+	// ignore machines not in pending state
```
ncdc: We said in the proposal that providers should not look at machine phase to determine what to do. What you can do here, instead, is ignore machines that have a non-nil spec.bootstrap.data value.

controllers/kubeadmconfig_controller.go
```diff
@@ -79,6 +78,11 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+	if machine.Status.Phase != v1alpha2.MachinePhaseStringPtr(v1alpha2.MachinePhasePending) {
```
detiber: We shouldn't be using the `Phase` for determining state, rather we should be using that relevant State/Ready fields. `Phase` should only be used for conveying information back to users.

controllers/kubeadmconfig_controller.go
ncdc: Or you could look at machine.status.bootstrapReady.

chuckha: ah...ok...yes, this is where we look at the real data instead of the field that is generated by the real data. Right. I remember that discussion now.

chuckha: I suspect that many providers will simply read the phase since it is available...

maybe the machine should do the reading and give me the phase, like

```
machine.Phase() == Pending
```

which looks directly at the data.

Thoughts @ncdc?

config/rbac/role.yaml
```diff
@@ -27,7 +27,7 @@ rules:
   - patch
   - update
 - apiGroups:
-  - cluster.sigs.k8s.io
+  - cluster.x-k8s.io
```
detiber: Should this be `bootstrap.cluster.x-k8s.io`?

chuckha: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/56/files/f94104b725fcd467b7ef495143db19abac482c58#diff-0406a3a4540c29330563040377417134R10

No, that is above. This is RBAC for the cluster objects.

chuckha: Which be scoped down in a future PR since we no longer need the watch on Machines

chuckha:
superseded by #62

PR #61, opened 2019-08-01
Use v1alpha2 api version for less user confusion

chuckha:
This also fixes a bad json field name in the Spec

/cc @amy 
/assign @detiber @ncdc 

This PR moves the apiversion to v1alpha2 and fixes a single typo in the Ready json field name.

```release-note
None
```


config/samples/kubeadm_v1alpha1_kubeadmconfig.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: bootstrap.cluster.sigs.k8s.io/v1alpha1
+apiVersion: bootstrap.cluster.sigs.k8s.io/v1alpha2
```
detiber: nit (for followup): We should probably add `config/samples` to .gitignore and remove these samples since they are not valid configs and just the autogenerated files from kubebuilder.

chuckha: 💯 i can do that in this pr

PR #67, opened 2019-08-05
Add AdditionalUserDataFiles field

### LINKED ISSUE #57: Add AdditionalUserDataFiles field
/kind feature

CAPA v0.3.x has a field called `AdditionalUserDataFiles` that looks like this:
```go
// AdditionalUserDataFiles specifies extra files to be passed to user_data upon creation.
// +optional
AdditionalUserDataFiles []userdata.Files `json:"additionalUserDataFiles,omitempty"`
```

We should add this for feature parity. CAPA also has this at the cluster scope too, to apply to all machines, in addition to at the machine scope. When we do #46, that should have an `AdditionalUserDataFiles` field too.

ncdc:
Add AdditionalUserDataFiles field to allow users to write arbitrary files to instances during cloud-init.

Fixes #57

**Release note**:

```release-note
NONE
```


api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -26,6 +27,9 @@ type KubeadmConfigSpec struct {
 	ClusterConfiguration kubeadmv1beta1.ClusterConfiguration `json:"clusterConfiguration"`
 	InitConfiguration    kubeadmv1beta1.InitConfiguration    `json:"initConfiguration,omitempty"`
 	JoinConfiguration    kubeadmv1beta1.JoinConfiguration    `json:"joinConfiguration,omitempty"`
+	// AdditionalUserDataFiles specifies extra files to be passed to user_data upon creation.
+	// +optional
+	AdditionalUserDataFiles []cloudinit.Files `json:"additionalUserDataFiles,omitempty"`
```
detiber: I would expect to also see an update to the generated CRDs for this update.

Should we expose types from the `cloudinit` package here? I would expect others importing our types wouldn't need to also import part of the implementation.

ncdc: I'll get the CRD yaml updated (CI should fail if that's not the case!)

In CAPA we were exposing `userdata.Files` but I see your point. I'll see about an update to correct it.

ncdc: @detiber updated, ptal

api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -61,3 +64,18 @@ type KubeadmConfigList struct {
...
+// Files defines the input for generating write_files in cloud-init.
+type Files struct {
+	// Path specifies the full path on disk where to store the file.
+	Path string `json:"path"`
+
+	// Owner specifies the ownership of the file, e.g. "root:root".
+	Owner string `json:"owner"`
+
+	// Permissions specifies the permissions to assign to the file, e.g. "0640".
+	Permissions string `json:"permissions"`
```
detiber: Permissions should probably be optional, since it isn't required for the cloud-init module.

api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -61,3 +64,18 @@ type KubeadmConfigList struct {
...
 	SchemeBuilder.Register(&KubeadmConfig{}, &KubeadmConfigList{})
 }
+
+// Files defines the input for generating write_files in cloud-init.
+type Files struct {
+	// Path specifies the full path on disk where to store the file.
+	Path string `json:"path"`
+
+	// Owner specifies the ownership of the file, e.g. "root:root".
+	Owner string `json:"owner"`
```
detiber: Owner should likely also be optional, since it isn't required for cloud-init.

api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -61,3 +64,18 @@ type KubeadmConfigList struct {
...
+
+	// Permissions specifies the permissions to assign to the file, e.g. "0640".
+	Permissions string `json:"permissions"`
+
+	// Content is the actual content of the file.
+	Content string `json:"content"`
+}
```
detiber: There are a couple other optional fields provided by cloud-init, should we also expose those?

`encoding` and `append` are the ones I'm thinking of and I'm not completely convinced that we necessarily *should* expose them.

ncdc: How about we leave them out for now, and we can add them later if needed?

api/v1alpha2/kubeadmbootstrapconfig_types.go
ncdc: Fixed, PTAL

api/v1alpha2/kubeadmbootstrapconfig_types.go
ncdc: Fixed, PTAL

api/v1alpha2/kubeadmbootstrapconfig_types.go
detiber: The data type lgtm, but I suspect that the template will need to be updated here: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/64d6676d48aab6ae55c36435176f147e03ea8a7a/cloudinit/files.go#L24-L25 as well.

ncdc: Good point. Working on fixing that now.

ncdc:
@detiber I updated the template logic. It took way too long to figure out how to omit empty lines when owner and permissions where empty! https://play.golang.org/p/AhjuCJrDC9u

chuckha:
This looks good
/approve

/assign @detiber 
for lgtm

ncdc:
I need to squash before this merges, FYI

chuckha:
@ncdc the approve label should stick around after changes to the PR, it's the LGTM that goes away, right?

ncdc:
@chuckha correct

chuckha:
Then I'm not really sure why you mention you have to squash before you merge? Makes sense if I was leaving an `/lgtm` label on here, but this was the approve label so @detiber could `/lgtm` later after the squash.

Would you prefer a different workflow?

ncdc:
That was more an indication to Jason not to remove the hold until I had squashed

detiber:
lgtm once squashed

ncdc:
Squashed

PR #76, opened 2019-08-05
Update to controller-tools v0.2.0-beta.5

### LINKED ISSUE #68: Update controller-tools to v2 beta5
/kind feature

Update controller-tools to v2 beta5. See https://github.com/kubernetes-sigs/cluster-api/pull/1224 as an example.

ncdc:

Fixes #68

This also adds a new hack/tools go mod.

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -85,4 +85,4 @@ docker-push:
 # Marking this as PHONY so it will build every time (so you always have the correct version locally)
 .PHONY: $(CONTROLLER_GEN)
 $(CONTROLLER_GEN):
-	go build -o $(CONTROLLER_GEN) sigs.k8s.io/controller-tools/cmd/controller-gen
+	cd hack/tools && go build -o $(CONTROLLER_GEN) sigs.k8s.io/controller-tools/cmd/controller-gen
```
detiber: Wouldn't this give us a potential diff between the vendored version of controller-tools for project and for the controller generation here?

chuckha: bin ends up in hack/tools/, not top level directory. I ended up make not able to find `bin/controller-gen`

chuckha: I would recommend this change:

```
CONTROLLER_GEN := hack/tools/bin/controller-gen
```

chuckha: wait no this is completely wrong

chuckha: Updated recommendation:

```
TOOLS_DIR := hack/tools
CONTROLLER_GEN := bin/controller-gen
CONTROLLER_GEN_PATH := $(TOOLS_DIR)/$(CONTROLLER_GEN)
```

```
# Build controller-gen
$(CONTROLLER_GEN_PATH): $(TOOLS_DIR)/go.mod
	cd $(TOOLS_DIR) && go build -o $(CONTROLLER_GEN) sigs.k8s.io/controller-tools/cmd/controller-gen
```

I'm kind of making stuff up here with removing the .PHONY but idk, might be cool?

Also changing all the CONTROLLER_GEN references to CONROLLER_GEN_PATH

ncdc: Updated with a slight variation on your suggestion - thanks @chuckha. PTAL!

Makefile
```diff
@@ -82,7 +84,5 @@ docker-push:
 	docker push ${MANAGER_IMAGE}
 
 # Build controller-gen
-# Marking this as PHONY so it will build every time (so you always have the correct version locally)
-.PHONY: $(CONTROLLER_GEN)
-$(CONTROLLER_GEN):
-	go build -o $(CONTROLLER_GEN) sigs.k8s.io/controller-tools/cmd/controller-gen
+$(CONTROLLER_GEN): $(TOOLS_DIR)/go.mod
```
detiber: I still worry that we have potentially reintroduced drift here between the version of controller-gen we are using here and the version of controller-tools we are vendoring for the controller.

BenTheElder: hmm, maybe pin both with a replace via some module tidy script?

ncdc: @detiber it's not vendored anywhere else. It's only used here. I need to regen the main go.sum to make it go away.

ncdc: Updated - ptal

ncdc:
Yay CI passed 💚

Makefile
chuckha: I think the concern is that controller-runtime needs to match controller-tools?

chuckha:
Looks good to me, I don't know enough about the version dependency between `controller-runtime` and `controller-tools`

Do those have to match?

detiber:
> Looks good to me, I don't know enough about the version dependency between controller-runtime and controller-tools

> Do those have to match?

Not necessarily, but they should be kept compatible. That is outside the scope of this PR, though :)

ncdc:
I'm not sure how strict the relationship is between controller-tools and controller-runtime. Right now, for example. controller-tools is at beta.5 while controller-runtime is still at beta.4.

chuckha:
In that case, I think this is good to go

/approve

PR #77, opened 2019-08-05
Verify manifests script

### LINKED ISSUE #73: CI should fail if CRD manifests are out of date
/kind bug

1. Modify a field in a CRD, such as changing the json tag
1. Run `hack/verify-all.sh`
1. It passes

1. It should fail because the generated manifests are out of date

**Environment:**

- Cluster API version:
- Cluster API Bootstrap Provider Kubeadm version:
- Cluster API Infrastructure Provider version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

amy:
- Adds verify manifests script

Fixes #73

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,12 @@ TOOLS_DIR := hack/tools
 CONTROLLER_GEN_BIN := bin/controller-gen
 CONTROLLER_GEN := $(TOOLS_DIR)/$(CONTROLLER_GEN_BIN)
 
+# Allow overriding manifest generation destination directory
+MANIFEST_ROOT ?= "config"
+CRD_ROOT ?= "$(MANIFEST_ROOT)/crd/bases"
+WEBHOOK_ROOT ?= "$(MANIFEST_ROOT)/webhook"
+RBAC_ROOT ?= "$(MANIFEST_ROOT)/rbac"
+
 # Active module mode, as we use go modules to manage dependencies
 export GO111MODULE=on
 
@@ -54,7 +60,7 @@ deploy: manifests
 # Generate manifests e.g. CRD, RBAC etc.
 .PHONY: manifests
 manifests: $(CONTROLLER_GEN)
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:dir=$(CRD_ROOT) output:webhook:dir=$(WEBHOOK_ROOT) output:rbac:dir=$(RBAC_ROOT)
 
 # Run go fmt against code
 .PHONY: fmt
--- a/hack/verify-manifests.sh
+++ b/hack/verify-manifests.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -53,8 +59,8 @@ deploy: manifests
 
 # Generate manifests e.g. CRD, RBAC etc.
 .PHONY: manifests
-manifests: $(CONTROLLER_GEN)
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+manifests: controller-gen
```
ncdc: Revert please and I'm guessing some of the other diffs in the PR might end up getting undone as well (different controller-gen version).

hack/verify-manifests.sh
```diff
@@ -0,0 +1,42 @@
...
+    rm -rf "${TMP_ROOT}"
+}
+trap "cleanup" EXIT SIGINT
+cleanup
+
+MANIFEST_ROOT=$TMP_ROOT make manifests
+
+echo "diffing ${DIFFROOT} against freshly generated manifests"
+ret=0
+git diff --no-index --diff-filter=MD ${TMP_ROOT} ${DIFFROOT} > /dev/null 2>&1 || ret=$?
```
amy: I am using `git diff` rather than `diff` here because the output is cleaner. I can't get `diff` to ignore files not present in the `TMP_ROOT`. There are many files in `DIFFROOT` that don't seem to be generated files.

ncdc: `git diff` will not see newly created files, so this test won't be 100% sufficient

amy: This test...
- generates expected manifest files in `_tmp`
- `git diff` will...
  - see if `/config` has all the files generated in `_tmp`
  - diff the matching `/config` and `_tmp` files

In which situation do we need to see newly created files in `config`.

amy: This test will...
- run `make manifests` and output the results into `_tmp`
- `git diff` will...
  - see if there exists files in `_tmp` that don't exist in `config`
  - see if the matching files in `_tmp` and `config` are the same

If by "newly created files" you mean files that exist in `config` and not in `_tmp`, there seemed to be many files in `config` that weren't generated.

Can you elaborate.

ncdc: Nevermind, I didn't see this was using `--no-index`. My bad!

amy: dope. Thanks for reviewing!

ncdc: Instead of redirecting to `/dev/null`, could you use the `--stat` flag to at least print a list of the files that it thinks are out of date?

amy: sure

ncdc:
@amy now that you have the file diff 😄:
```
diffing hack/../config against freshly generated manifests
 .../bootstrap.cluster.x-k8s.io_kubeadmconfigs.yaml | 26 ++++++++++++++++++++++
 1 file changed, 26 insertions(+)
Manifests in hack/../config are out of date. Please run `make manifests`
```

Can you see if something truly is out of date?

amy:
Yah. Just did another rebase. Looks like running `make manifests` again removes this:
```
-            additionalUserDataFiles:
-              description: AdditionalUserDataFiles specifies extra files to be passed
-                to user_data upon creation.
-              items:
-                description: Files defines the input for generating write_files in
-                  cloud-init.
-                properties:
-                  content:
-                    description: Content is the actual content of the file.
-                    type: string
-                  owner:
-                    description: Owner specifies the ownership of the file, e.g. "root:root".
-                    type: string
-                  path:
-                    description: Path specifies the full path on disk where to store
-                      the file.
-                    type: string
-                  permissions:
-                    description: Permissions specifies the permissions to assign to
-                      the file, e.g. "0640".
-                    type: string
-                required:
-                - content
-                - path
-                type: object
-              type: array
```

ncdc:
ugh, apparently yesterday's rebase of #42 managed to eliminate part/all of #57 😦 cc @chuckha

ncdc:
I've fixed master so now we shouldn't fail manifest verification (#89)

chuckha:
oh nooo, did i hose a rebase -_-

PR #69, opened 2019-08-05
Switch from Update() to Patch()

ncdc:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #63 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note
NONE
```

PR #82, opened 2019-08-06
Reconciler loop

fabriziopandini:
@ncdc @chuckha 
disclaimer: I'm still in learning mode.

This PR is an attempt to get the reconciler loop properly manage the creation of the init control plane. Main points I want to discuss with you:
1. While `cluster.x-k8s.io/infrastructure-ready` is not set, everything should wait
1. If `cluster.x-k8s.io/infrastructure-ready==true`, while `cluster.x-k8s.io/control-plane-ready` is not set
  1. we should ensure only one config/one machine is set for acting as init control plane; this requires the init-lock mechanism to be in place
  2. everything else should wait
1. if `cluster.x-k8s.io/infrastructure-ready==true` and `cluster.x-k8s.io/control-plane-ready==true the reconciler loop should handle join control plane and join workers

### DIFF
--- a/cloudinit/controlplane_join.go
+++ b/cloudinit/controlplane_join.go
@@ -46,6 +46,11 @@ type ControlPlaneJoinInput struct {
 // NewJoinControlPlane returns the user data string to be used on a new control plane instance.
 func NewJoinControlPlane(input *ControlPlaneJoinInput) ([]byte, error) {
 	input.Header = cloudConfigHeader
+	if err := input.Certificates.validate(); err != nil {
+		return nil, errors.Wrapf(err, "ControlPlaneInput is invalid")
+	}
+
+	input.WriteFiles = certificatesToFiles(input.Certificates)
 	input.WriteFiles = append(input.WriteFiles, input.AdditionalFiles...)
 	userData, err := generate("JoinControlplane", controlPlaneJoinCloudInit, input)
 	if err != nil {
--- a/cloudinit/controlplane_join.go
+++ b/cloudinit/controlplane_join.go
@@ -39,19 +39,13 @@ type ControlPlaneJoinInput struct {
 	BaseUserData
 	Certificates
 
-	BootstrapToken      string
-	ControlPlaneAddress string
-	JoinConfiguration   string
+	BootstrapToken    string
+	JoinConfiguration string
 }
 
 // NewJoinControlPlane returns the user data string to be used on a new control plane instance.
 func NewJoinControlPlane(input *ControlPlaneJoinInput) ([]byte, error) {
 	input.Header = cloudConfigHeader
-	if err := input.Certificates.validate(); err != nil {
-		return nil, errors.Wrapf(err, "ControlPlaneInput is invalid")
-	}
-
-	input.WriteFiles = certificatesToFiles(input.Certificates)
 	input.WriteFiles = append(input.WriteFiles, input.AdditionalFiles...)
 	userData, err := generate("JoinControlplane", controlPlaneJoinCloudInit, input)
 	if err != nil {
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -26,22 +26,24 @@ import (
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kubeadmv1alpha2 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
-	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/cloudinit"
-	"sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
+	capiv1alpha2 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
 	"sigs.k8s.io/cluster-api/pkg/util"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	cabpkv1alpha2 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/cloudinit"
+	kubeadmv1beta1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
 )
 
 var (
-	machineKind = v1alpha2.SchemeGroupVersion.WithKind("Machine")
+	machineKind = capiv1alpha2.SchemeGroupVersion.WithKind("Machine")
 )
 
 const (
 	// InfrastructureReadyAnnotationKey identifies when the infrastructure is ready for use such as joining new nodes.
 	// TODO move this into cluster-api to be imported by providers
-	InfrastructureReadyAnnotationKey = "cluster.x-k8s.io/infrastructure-ready"
+	ControlPlaneReadyAnnotationKey = "cluster.x-k8s.io/control-plane-ready"
 )
 
 // KubeadmConfigReconciler reconciles a KubeadmConfig object
@@ -59,7 +61,7 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 	ctx := context.Background()
 	log := r.Log.WithValues("kubeadmconfig", req.NamespacedName)
 
-	config := kubeadmv1alpha2.KubeadmConfig{}
+	config := cabpkv1alpha2.KubeadmConfig{}
 	if err := r.Get(ctx, req.NamespacedName, &config); err != nil {
 		if apierrors.IsNotFound(err) {
 			return ctrl.Result{}, nil
@@ -75,6 +77,7 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 	}
 
 	// Find the owner reference
+	// The cluster-api machine controller set this value.
 	var machineRef *v1.OwnerReference
 	for _, ref := range config.OwnerReferences {
 		if ref.Kind == machineKind.Kind && ref.APIVersion == machineKind.GroupVersion().String() {
@@ -88,7 +91,7 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 	}
 
 	// Get the machine
-	machine := &v1alpha2.Machine{}
+	machine := &capiv1alpha2.Machine{}
 	machineKey := client.ObjectKey{
 		Namespace: req.Namespace,
 		Name:      machineRef.Name,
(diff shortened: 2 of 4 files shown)


cloudinit/controlplane_init.go
```diff
@@ -24,9 +24,9 @@ const (
     permissions: '0640'
     content: |
       ---
-{{.ClusterConfiguration | Indent 6}}
+{{.ClusterConfiguration}}
```
fabriziopandini: `| Indent 6` is a jinja templating things... removed

controllers/kubeadmconfig_controller.go
```diff
@@ -166,13 +131,83 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+			})
+			if err != nil {
+				log.Error(err, "failed to generate cloud init for bootstrap control plane")
+				return ctrl.Result{}, err
+			}
+
+			config.Status.BootstrapData = cloudInitData
+			config.Status.Ready = true
+
+			if err := r.Update(ctx, &config); err != nil {
```
fabriziopandini: Update or Patch?

cloudinit/controlplane_init.go
ncdc: It was a go template function we wrote in CAPA. I'd recommend bringing it over. Otherwise I don't think it will be valid yaml?

controllers/kubeadmconfig_controller.go
```diff
@@ -42,6 +42,7 @@ const (
 	// InfrastructureReadyAnnotationKey identifies when the infrastructure is ready for use such as joining new nodes.
 	// TODO move this into cluster-api to be imported by providers
 	InfrastructureReadyAnnotationKey = "cluster.x-k8s.io/infrastructure-ready"
```
ncdc: We should be able to remove this. In v1alpha2, we have `cluster.status.infrastructureReady` as a `bool`

controllers/kubeadmconfig_controller.go
```diff
@@ -119,43 +120,7 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
-			log.Error(err, "failed to update config")
-			return ctrl.Result{}, err
-		}
-		return ctrl.Result{}, nil
-	}
-
-	// Check for control-plane ready. If it's not ready then we will requeue the machine until it is.
+	// Check for infrastructure ready. If it's not ready then we will requeue the machine until it is.
 	// The infrastructure provider *must* set this value to use this bootstrap provider.
 	if cluster.Annotations == nil {
```
ncdc: We maybe can get rid of this check entirely. I'd say it's safe to generate the bootstrap data regardless of the cluster infrastructure's status. We can move this burden to each infrastructure provider - they should not proceed to create instances from Machines until the Cluster infra is ready. WDYT?

controllers/kubeadmconfig_controller.go
```diff
@@ -166,13 +131,83 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 	}
 
+	// Store Config's state, pre-modifications, to allow patching
+	patchConfig := client.MergeFrom(config.DeepCopy())
+
+	// Check for control plane ready. If it's not ready then we will requeue the machine until it is.
+	// The infrastructure provider *must* set this value to use this bootstrap provider.
```
ncdc: This is set by cluster-api, not an infra provider. See https://github.com/kubernetes-sigs/cluster-api/blob/62582da077611649856cb2a88ec73c957e8569f5/pkg/controller/machine/machine_controller_phases.go#L264-L288

controllers/kubeadmconfig_controller.go
ncdc: Patch please

chuckha: I think there was a rebase gone awry somewhere. This should use patch.

go.sum
```diff
@@ -197,7 +197,7 @@ golang.org/x/tools v0.0.0-20190114222345-bf090417da8b/go.mod h1:n7NCudcB/nEzxVGm
 golang.org/x/tools v0.0.0-20190226205152-f727befe758c/go.mod h1:9Yl7xja0Znq3iFh3HoIrodX9oNMXvdceNzlUR8zjMvY=
 golang.org/x/tools v0.0.0-20190312170243-e65039ee4138/go.mod h1:LCzVGOaR6xXOjkQ3onu1FJEFr0SW1gC7cKk1uF8kGRs=
 golang.org/x/tools v0.0.0-20190501045030-23463209683d/go.mod h1:RgjU9mgBXZiqYHBnxXauZ1Gv1EHHAz9KjViQ78xBX0Q=
-gomodules.xyz/jsonpatch/v2 v2.0.0 h1:lHNQverf0+Gm1TbSbVIDWVXOhZ2FpZopxRqpr2uIjs4=
+gomodules.xyz/jsonpatch/v2 v2.0.0 h1:OyHbl+7IOECpPKfVK42oFr6N7+Y2dR+Jsb/IiDV3hOo=
```
chuckha: This isn't right, please generate the go.sum using GOPROXY set to some proxy. One such proxy is https://proxy.golang.org

controllers/kubeadmconfig_controller.go
chuckha: this is my comment, yeah, it was set in v1alpha1 by the infra, glad to see it moved to capi

controllers/kubeadmconfig_controller.go
fabriziopandini: I'm assuming we should wait for the load balancer to get created in order to use it for `ClusterConfiguration.ControlPlaneEndpoint`

controllers/kubeadmconfig_controller.go
chuckha: oh excellent

cloudinit/controlplane_init.go
fabriziopandini: Gotcha.
The `Indent` go template func is already there, but it expects `int, string`, while `ClusterConfiguration` ATM is `[]byte`. Fix in flight

controllers/kubeadmconfig_controller.go
ncdc: I just opened #85 but you raise a good point - let's continue the conversation there.

ncdc: Nevermind - we decided to keep the cluster status infraReady check

controllers/kubeadmconfig_controller.go
fabriziopandini: @ncdc @chuckha There are some weird problem after this change.
It seems that every operation with `k8sClient` implemented in tthe test loses GVK, and this leads `r.Status().Patch` to fail with "the server could not find the requested resource (patch kubeadmconfigs.bootstrap.cluster.x-k8s.io control-plane-config)". 
Any suggestion?

controllers/kubeadmconfig_controller.go
fabriziopandini: fixed

controllers/kubeadmconfig_controller.go
ncdc: Known issue, 1 sec

ncdc: Copy what's in here: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/69/files

fabriziopandini: @ncdc the problem is a little bit different
when I call `k8sClient.Create(context.Background(), config)` in tests, it seems the GVK gets stripped from config before being stored in etcd, or at least this is what I assume when I check config after calling of `k8sClient.Get(ctx, controlplaneConfigKey, config)`

ncdc: When apimachinery decodes into a typed struct, the decoder strips the TypeMeta from the object (the theory at the time being that because it was a typed object, you knew its API version, group, and kind). There's a WIP PR to k/k to fix this.

This sounds like the same issue - right?

fabriziopandini:
@ncdc @chuckha I have pushed an update on this PR.

apart rebase + address comments, major changes are:
- tolerate one of InitConfiguration or ClusterConfiguration missing (same as kubeadm)
- if defined, inject cluster.Status.APIEndpoints into ClusterConfiguration for the init control plane
- fail if control plane is ready, but cluster.Status.APIEndpoints is not set
- inject cluster.Status.APIEndpoints into JoinConfiguration
- fail if joining a control plane machine, but the kubeadm config does not join as a control plane
- lot of unit tests

controllers/kubeadmconfig_controller.go
```diff
@@ -104,41 +107,107 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+
+		if config.Spec.ClusterConfiguration == nil {
+			config.Spec.ClusterConfiguration = &kubeadmv1beta1.ClusterConfiguration{
+				TypeMeta: v1.TypeMeta{
+					APIVersion: "kubeadm.k8s.io/v1beta1",
+					Kind:       "ClusterConfiguration",
+				},
+			}
+		}
+		// If there is a control plane endpoint defined at cluster in cluster status, use it as a control plane endpoint for the K8s cluster
```
detiber: This comment should also indicate that we are only using the first one defined if there are multiple defined.

controllers/kubeadmconfig_controller.go
```diff
@@ -104,41 +107,107 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
 		if err != nil {
 			log.Error(err, "failed to marshal cluster configuration")
 			return ctrl.Result{}, err
 		}
 
+		//TODO(fp) check what is the expected flow for certificates
+		certificates, _ := NewCertificates()
+
 		cloudInitData, err := cloudinit.NewInitControlPlane(&cloudinit.ControlPlaneInput{
+			//TODO(fp) if defined, cluster.Status.APIEndpoints[0] should override ClusterConfiguration.ControlPlaneEndpoint
```
detiber: Based on lines 184-187, I think this comment can now be removed.

controllers/kubeadmconfig_controller.go
```diff
@@ -148,32 +217,51 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+		return ctrl.Result{}, errors.New("Control plane already exists for the cluster but cluster.Status.APIEndpoints are not set")
+	}
+
+	// Injects the controlplane endpoint address
+	// NB. this assumes nodes using token discovery
+	if config.Spec.JoinConfiguration.Discovery.BootstrapToken == nil {
+		config.Spec.JoinConfiguration.Discovery.BootstrapToken = &kubeadmv1beta1.BootstrapTokenDiscovery{}
+	}
+
+	config.Spec.JoinConfiguration.Discovery.BootstrapToken.APIServerEndpoint = fmt.Sprintf("https://%s:%d", cluster.Status.APIEndpoints[0].Host, cluster.Status.APIEndpoints[0].Port)
```
detiber: Is the bootstrap token being generated elsewhere? Or is that still an outstanding todo?

cloudinit/controlplane_join.go
```diff
@@ -39,19 +39,13 @@ type ControlPlaneJoinInput struct {
...
 }
 
 // NewJoinControlPlane returns the user data string to be used on a new control plane instance.
 func NewJoinControlPlane(input *ControlPlaneJoinInput) ([]byte, error) {
 	input.Header = cloudConfigHeader
-	if err := input.Certificates.validate(); err != nil {
-		return nil, errors.Wrapf(err, "ControlPlaneInput is invalid")
-	}
-
-	input.WriteFiles = certificatesToFiles(input.Certificates)
```
detiber: Are we intending to drop support for k8s/kubeadm v1.13 here? Otherwise we'd still need to write the certificates to disk for a control plane join.

controllers/kubeadmconfig_controller.go
fabriziopandini: I assume this should be generated elsewhere because I don't see how CABPK can generate a bootstrap token and ensure it gets passed to all the machine in the cluster, but I can be totally wrong.... @ncdc @chuckha suggestions?

ncdc: In CAPA v1alpha1, we generate a bootstrap token secret, store it in the apiserver, and then plumb it through to the cloud-init data for each machine. We can do the same for CABPK.

cloudinit/controlplane_join.go
fabriziopandini: good point. it is still not clear to me what is the expected flow for certificates, I have to sync with @ncdc @chuckha ...

ncdc: I'm not up to speed on <=1.13 vs >= 1.14 & kubeadm & certs. Can someone please fill me in?

detiber: v1.14+ it is possible to do a kubeadm join without manually copying the certificates using by using a encrypted and timeboxed secret

controllers/kubeadmconfig_controller.go
chuckha: It's not implemented yet, but the idea is exactly as what @ncdc stated

cloudinit/controlplane_join.go
chuckha: My impression of using the --certificate-upload flag was that it was for human users and that automating it would be more of a pain than automating the cert creation.

My expected workflow was => on create or join look for certificate secrets. If they exist then write them to disk (/etc/kubernetes/pki). If they don't exist then generate the CAs and upload those to secrets (should only happen with first control plane). It doesn't feel very secure since secrets aren't particularly secret.

detiber: @chuckha since we already have the secrets we could basically replicate the --certificate-upload process as part of the bootstrapping as we do today for the bootstrap tokens.

controllers/kubeadmconfig_controller.go
detiber: As an aside, I'm wondering if when we implement it here we should make sure that we shorten the lifetime of the bootstrap token.

controllers/kubeadmconfig_controller.go
fabriziopandini: ok, for the time being, I commented on the behavioral tests that was failing;
considering that everything is covered by unit tests, IMO this can be addressed in follow up PRs when the root of this issue is fixed in k/k

cloudinit/controlplane_join.go
fabriziopandini: kk. I restored the deleted lines and isolated certificate generation in a separated func, so we can plugin the certificate flow described above in a follow-up PR

controllers/kubeadmconfig_controller.go
```diff
@@ -26,22 +26,24 @@ import (
...
+	kubeadmv1beta1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
 )
 
 var (
-	machineKind = v1alpha2.SchemeGroupVersion.WithKind("Machine")
+	machineKind = capiv1alpha2.SchemeGroupVersion.WithKind("Machine")
 )
 
 const (
 	// InfrastructureReadyAnnotationKey identifies when the infrastructure is ready for use such as joining new nodes.
```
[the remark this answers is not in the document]
fabriziopandini: it should be updated, not removed

fabriziopandini:
@chuckha @ncdc @detiber thanks for the feedback!
everything should be addressed, and if it is ok for you we can merge this and move on next iterations for addressing:
- token management
- certificate management
- init-lock (as a safeguard for more control-plane nodes configured for init)
- behavioral test (blocked by failure on Patch due to API machinery losing gvk)

Issue #85, opened 2019-08-06
Should we honor or ignore cluster.spec.infrastructureReady?


ncdc:
In CAPA for v1alpha1, we added code that prevented any Machine instances (ec2 instances) from being created if the cluster infrastructure (VPC, security groups, etc.) wasn't ready yet.

Now that generating bootstrap data is isolated from provisioning Machine infrastructure, I'm wondering if it would be ok to remove the logic in this repo that prevents generating bootstrap data if `cluster.status.infrastructureReady` is `false`. The infrastructure providers _should_ already be gating Machine infrastructure provisioning based on `cluster.status.infrastructureReady`. WDYT?

ncdc:
@chuckha @fabriziopandini

detiber:
@ncdc I'm not entirely sure, since in the case of AWS, the load balancer is created as part of the infrastructure and would be needed for generating the bootstrap data.

ncdc:
@detiber true (and @fabriziopandini mentioned it too in #82)... but we could have this repo require `cluster.status.apiEndpoints` to be valid before continuing (or have the user fill it in manually in the kubeadm config).

detiber:
@ncdc The potential problem would be is that we would just know that the endpoints are set and not that the shared infrastructure is actually ready to be consumed at that point. I think having an explicit InfrastructureReady field allows the cluster infrastructure provider to have a clear marker to signal to the bootstrapper/machine controller that it is ok to proceed is better than implicitly looking for endpoints.

There are also cases today where the API endpoints are not set by the cluster infrastructure and are instead set by the first Machine that is bootstrapped, which would present issues if we are using it to gate machine bootstrapping.

ncdc:
`cluster.spec.infrastructureRef` is optional, so we may need to tell users who leave that unset to patch the status to set `infrastructureReady` to true.

detiber:
yuck, should we provide an optional `noop` cluster infrastructure provider that would just set `infrastructureReady=true` for those use cases?

ncdc:
If we add the `spec.defaultInfrastructureProvider` field you suggested, that could work, although it would end up creating 1 NoopCluster for each Cluster, which seems unnecessary.

For now, I say we honor and require `cluster.status.infrastructureReady` based on the arguments made above, and we can deal with optional infraRef elsewhere. WDYT?

detiber:
sgtm

PR #92, opened 2019-08-06
Restore control plane init locker

fabriziopandini:
As discussed in https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/82, the control plane init locker is required in order to ensure only a KubeadmConfig/Machine in the cluster is configured as a init node. 

NB. This is a safeguard that prevents errors in case more that one KubeadmConfig/Machine is configured with ClusterConfiguration/InitConfiguration

/assign @chuckha

### DIFF
--- a/controllers/control_plane_init_locker.go
+++ b/controllers/control_plane_init_locker.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	apicorev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	clusterv2 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
+)
+
+// ControlPlaneInitLocker provides a locking mechanism for cluster initialization.
+type ControlPlaneInitLocker interface {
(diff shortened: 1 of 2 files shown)


detiber:
I didn't see the discussion in #82, but how would we be able to join the a second Machine that was defined with ClusterConfiguration/InitConfiguration? Or would we just block indefinitely?

chuckha:
I don't see the discussion in #82, can you please link to the comment @fabriziopandini ?

@detiber it should just be a join configuration even for a control plane, there shouldn't be a second machine with ClusterConfiguration/InitConfiguration. That would end up making an entirely new cluster.

detiber:
@chuckha right, that is why I was confused by the need for the locking.

fabriziopandini:
@detiber @chuckha 

It is in PR main comment (AI for me: next time make items for discussion more visible, might be opening separated issues like https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/85)

> ....
> 2. If `cluster.x-k8s.io/infrastructure-ready==true`, while `cluster.x-k8s.io/control-plane-ready` is not set
>     - we should ensure only one config/one machine is set for acting as init control plane; this requires the init-lock mechanism to be in place
>     - everything else should wait

The need for init-lock is only for preventing errors in case more that one KubeadmConfig/Machine is configured with ClusterConfiguration/InitConfiguration

> there shouldn't be a second machine with ClusterConfiguration/InitConfiguration

yes, but nothing prevents it to happen so IMO we should handle this case

chuckha:
There are a lot of user errors we can't/won't account for, but if we can at least prevent a user from shooting themselves in the foot that's ok with me.

What is the plan around how the controller will behave if a second bootstrap control plane is found for a given cluster?

ncdc:
Do we need to consider adding errorReason and errorMessage to KubeadmConfigStatus (and all other bootstrap CRD statuses)? cc @vincepri

detiber:
I think it would probably be a good idea to go ahead and add ErrorMessage/ErrorReason

fabriziopandini:
> What is the plan around how the controller will behave if a second bootstrap control plane is found for a given cluster?

IMO the controller should requeue the second bootstrap control plane and try to join it as soon as the controlplane is ready.

That means that if by chance the second bootstrap control plane has a JoinConfiguration, then it will join (InitConfiguration and Clusterconfiguration will be ignored at this point); If the second bootstrap control plane has not JoinConfiguration the controller will error out (and eventually update ErrorMessage/ErrorReason).

chuckha:
Sounds like we're in agreement that this is a case we should handle. I think @fabriziopandini has the right idea.

/lgtm

chuckha:
I will open an issue to add ErrorReason/ErrorMessage to the types

chuckha:
However I should mention that this may not make it into the first release since we're aiming for end of month and that is ok.

PR #95, opened 2019-08-08
Refactor: use Patch instead of Update to update a config

### LINKED ISSUE #94: All Updates should be Patches
/kind feature

Refactor the controller to only use Patch to update a config. Don't use update.

/priority important-soon
/milestone v0.1
/good-first-issue
/assign @joonas
/lifecycle active

SataQiu:
Refactor the controller to only use Patch to update a config. Don't use update.

Fixes #94

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -271,8 +271,8 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 
 		config.Status.BootstrapData = joinData
 		config.Status.Ready = true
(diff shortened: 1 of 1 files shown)


ykakarap:
Looks good.

I don't have permissions to approve through.

chuckha:
This is not quite right, the status should also be patched in these cases as well as handling the work around for the controller-runtime bug.

Please see https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/95/files#diff-3dec0e9530e56d497bc62428156004faR202-R216

PR #103, opened 2019-08-08
Set GOPROXY to default if it is not configured

joonas:

This is necessary since `go env GOPROXY` will return an empty string if the value is not set, thus setting GOPROXY to an empty string, which doesn't quite do what we want.

I've verified this locally by doing a fresh checkout without `GOPROXY` being set and running `make test`, which resulted in:
```
verifying gomodules.xyz/jsonpatch/v2@v2.0.0: checksum mismatch
	downloaded: h1:OyHbl+7IOECpPKfVK42oFr6N7+Y2dR+Jsb/IiDV3hOo=
	go.sum:     h1:lHNQverf0+Gm1TbSbVIDWVXOhZ2FpZopxRqpr2uIjs4=
make: *** [fmt] Error 1
```

After applying this change and running `make test` again things worked as expected.

**Release note**:
```release-note
NONE
```

/cc @chuckha @amy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 # Use GOPROXY environment variable if set
 GOPROXY := $(shell go env GOPROXY)
-GOPROXY ?= https://proxy.golang.org
+ifeq ($(GOPROXY),)
+GOPROXY := https://proxy.golang.org
+endif
 export GOPROXY
 
 REGISTRY ?= gcr.io/$(shell gcloud config get-value project)


ncdc:
/hold
we want to honor `go env GOPROXY` if we can

ncdc:
I have a clean checkout, I don't have GOPROXY set, and `make test` passes for me...

joonas:
> I have a clean checkout, I don't have GOPROXY set, and `make test` passes for me...

Just to check, did you run `go clean -modcache` before running `make test` to make sure you have an empty cache?

ncdc:
Doing that now. Will report back

ncdc:
Ok I get the same error with a clean modcache

joonas:
> Ok I get the same error with a clean modcache

Sorry I should've mentioned that being a pre-requisite originally on the ticket.

The goal is to honor `GOPROXY` if it's set, but set a reasonable default in case it's not in order to get around the error with `jsonpatch`.

ncdc:
How about this?
```
GOPROXY := $(shell go env GOPROXY)
ifeq ($(GOPROXY),)
GOPROXY := https://proxy.golang.org
endif
```

joonas:
Happy to change it to that, that's definitely more succinct

ncdc:
LGTM, please squash & thanks!
/approve

chuckha:
/lgtm

thanks @joonas

Issue #98, opened 2019-08-08
Create bootstrap tokens for joining nodes


fabriziopandini:
/kind feature

**Describe the solution you'd like**
As per [comment](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/82#discussion_r311600674), before joining a node the reconciler should create a new bootstrap token in the target cluster

**Anything else you would like to add:**
Relevant code in CAPAv1alpha1
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/release-0.3/pkg/cloud/aws/actuators/machine/actuator.go#L149
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/release-0.3/pkg/tokens/tokens.go

/assign
/lifecycle active

PR #105, opened 2019-08-09
Reconcile automatically create join token if necessary

fabriziopandini:
before joining a node the reconciler should create a new bootstrap token in the target cluster

**Which issue(s) this PR fixes**:
Fixes: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/98

**Release note**:
```release-note
NONE
```


fabriziopandini:
@chuckha @ncdc @detiber this is ready for review now

controllers/kubeadmconfig_controller.go
```diff
@@ -155,7 +157,7 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 		log.Info("Creating BootstrapData for the init control plane")
 
 		// get both of ClusterConfiguration and InitConfiguration strings to pass to the cloud init control plane generator
-		// kubeadm allows one this values to be empty; cabpk replace missing values with an empty config, so the cloud init generation
+		// kubeadm allows one this values to be empty; CAPBK replace missing values with an empty config, so the cloud init generation
```
chuckha:
```suggestion
		// kubeadm allows one this values to be empty; CABPK replace missing values with an empty config, so the cloud init generation
```

controllers/kubeadmconfig_controller.go
```diff
@@ -180,10 +182,12 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 				},
 			}
 		}
-		// If there is a control plane endpoint defined at cluster in cluster status, use it as a control plane endpoint for the K8s cluster
-		// NB. we are only using the first one defined if there are multiple defined.
-		if len(cluster.Status.APIEndpoints) > 0 {
+		// If there no ControlPlaneEndpoint defined in ClusterConfiguration but there are APIEndpoints defined at cluster level (e.g. the load balancer endpoint),
+		// then use cluster APIEndpoints as a control plane endpoint for the K8s cluster
+		if config.Spec.ClusterConfiguration.ControlPlaneEndpoint == "" && len(cluster.Status.APIEndpoints) > 0 {
+			// NB. CAPBK only uses the first APIServerEndpoint defined in cluster status if there are multiple defined.
```
chuckha:
```suggestion
			// NB. CABPK only uses the first APIServerEndpoint defined in cluster status if there are multiple defined.
```

controllers/kubeadmconfig_controller.go
```diff
@@ -300,6 +302,68 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 	return ctrl.Result{}, nil
 }
 
+// reconcileDiscovery ensure that joinConfiguration.Discovery is properly set for the joining node.
+// The implementation func respect user provided discovery configurations, but in case some of them are missing, a valid BootstrapToken object
+// is automatically injected into joinConfiguration.Discovery.
+// This allows to simplify configuration UX, by providing the option to delegate to CAPBK the configuration of kubeadm join discovery.
```
chuckha:
```suggestion
// This allows to simplify configuration UX, by providing the option to delegate to CABPK the configuration of kubeadm join discovery.
```

controllers/kubeadmconfig_controller.go
```diff
@@ -300,6 +302,68 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+
+	// if BootstrapToken already contains an APIServerEndpoint, respect it; otherwise inject the APIServerEndpoint endpoint defined in cluster status
+	//TODO(fp) might be we want to validate user provided APIServerEndpoint and warn/error if it doesn't match the api endpoint defined at cluster level
+	apiServerEndpoint := discovery.BootstrapToken.APIServerEndpoint
+	if apiServerEndpoint == "" {
+		if len(cluster.Status.APIEndpoints) == 0 {
+			return errors.New("Control plane already exists for the cluster but cluster.Status.APIEndpoints are not set")
+		}
+
+		// NB. CAPBK only uses the first APIServerEndpoint defined in cluster status if there are multiple defined.
```
chuckha:
```suggestion
		// NB. CABPK only uses the first APIServerEndpoint defined in cluster status if there are multiple defined.
```

controllers/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+	bootstraputil "k8s.io/cluster-bootstrap/token/util"
+	capiremote "sigs.k8s.io/cluster-api/pkg/controller/remote"
+)
+
+const (
+	defaultTokenTTL = 10 * time.Minute
+)
+
+// createTokenInRemoteCluster attempts to create a token with the given ID.
+func createTokenInRemoteCluster(remoteclient capiremote.ClusterClient) (string, error) {
```
chuckha: This interface is too wide, we only really need  to pass in an interface that can run `Create(*v1.Secret)` The caller can set up a remote client and pass in a creator. I would also recommend changing the name to `createToken` since the client is the thing that dictates which cluster this is created in. That detail should not be hard-coded.

Calling this function would look like this:

```
creator := remoteclient.CoreV1().Secrets(metav1.NamespaceSystem)
createToken(creator)
```

controllers/kubeadmconfig_controller.go
```diff
@@ -300,6 +302,68 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
 	return ctrl.Result{}, nil
 }
 
+// reconcileDiscovery ensure that joinConfiguration.Discovery is properly set for the joining node.
+// The implementation func respect user provided discovery configurations, but in case some of them are missing, a valid BootstrapToken object
+// is automatically injected into joinConfiguration.Discovery.
+// This allows to simplify configuration UX, by providing the option to delegate to CAPBK the configuration of kubeadm join discovery.
+func (r *KubeadmConfigReconciler) reconcileDiscovery(cluster *capiv1alpha2.Cluster, discovery *kubeadmv1beta1.Discovery, log logr.Logger) error {
```
chuckha: KubeadmConfigReconciler already has a `logr.Logger`. We don't need to pass one in here. But we should create a new one at the top of this function if some logging scope is needed.

controllers/kubeadmconfig_controller.go
```diff
@@ -300,6 +302,68 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
+		// NB. CAPBK only uses the first APIServerEndpoint defined in cluster status if there are multiple defined.
+		apiServerEndpoint = fmt.Sprintf("https://%s:%d", cluster.Status.APIEndpoints[0].Host, cluster.Status.APIEndpoints[0].Port)
+		discovery.BootstrapToken.APIServerEndpoint = apiServerEndpoint
+		log.Info("Altering JoinConfiguration.Discovery.BootstrapToken", "APIServerEndpoint", apiServerEndpoint)
+	}
+
+	// if BootstrapToken already contains a token, respect it; otherwise create a new bootstrap token for the node to join
+	if discovery.BootstrapToken.Token == "" {
+		log.Info("Creating a bootstrap token in the remote cluster")
+		if r.remoteClientFactory == nil {
```
chuckha: We cannot put a remote client factory on the reconciler as the reconciler is a singleton that needs to work across many clusters.

Instead, I suggest creating a client when necessary during every reconcile.

controllers/kubeadmconfig_controller.go
detiber:
```suggestion
		// kubeadm allows one of these values to be empty; CABPK replace missing values with an empty config, so the cloud init generation
```

controllers/kubeadmconfig_controller_test.go
```diff
@@ -18,19 +18,25 @@ package controllers
...
-	kubeadmv1beta1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
+	fakeclient "k8s.io/client-go/kubernetes/fake"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	restclient "k8s.io/client-go/rest"
 	capiv1alpha2 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
+	"sigs.k8s.io/cluster-api/pkg/controller/remote"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/runtime/log"
+
```
chuckha: across projects we've been breaking imports up into only two sections,

```
<stdlibrary alphabetized>

<not stdlibrary alphabetized>
```

Do you think we should start doing it more like

```
<stdlib alphabetized>

<third party alphabetized>

<this package alphabetized>
```
?

vincepri: I'd usually prefer the first, which is what `goimports` does by default

controllers/kubeadmconfig_controller.go
fabriziopandini: the factory is a function that gets a cluster as an input, so, it should be already designed to work across many clusters.

it is defined at reconciler level in order to get everything unit-testable

fabriziopandini:
@chuckha rebased + addressed the first round of feedback

controllers/kubeadmconfig_controller.go
```diff
@@ -263,6 +267,83 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+	if config.Spec.JoinConfiguration.Discovery.BootstrapToken == nil {
+		config.Spec.JoinConfiguration.Discovery.BootstrapToken = &kubeadmv1beta1.BootstrapTokenDiscovery{}
+	}
+
+	// if BootstrapToken already contains an APIServerEndpoint, respect it; otherwise inject the APIServerEndpoint endpoint defined in cluster status
+	//TODO(fp) might be we want to validate user provided APIServerEndpoint and warn/error if it doesn't match the api endpoint defined at cluster level
+	apiServerEndpoint := config.Spec.JoinConfiguration.Discovery.BootstrapToken.APIServerEndpoint
+	if apiServerEndpoint == "" {
+		if len(cluster.Status.APIEndpoints) == 0 {
+			return errors.New("Control plane already exists for the cluster but cluster.Status.APIEndpoints are not set")
```
chuckha: I would expect to return requeue after some duration instead of returning an error here. My thinking is that if the status.APIEndpoints doesn't exist then it means we're waiting on cluster-api to write that value.

controllers/kubeadmconfig_controller.go
```diff
@@ -263,6 +267,83 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+		}
+
+		// NB. CABPK only uses the first APIServerEndpoint defined in cluster status if there are multiple defined.
+		apiServerEndpoint = fmt.Sprintf("https://%s:%d", cluster.Status.APIEndpoints[0].Host, cluster.Status.APIEndpoints[0].Port)
```
chuckha: i know the controlPlaneEndpoint should *not* have `https://` prefixed. Is that different for the bootstrapToken apiserverendpoint?

controllers/kubeadmconfig_controller.go
```diff
@@ -263,6 +267,83 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+		log.Info("Altering JoinConfiguration.Discovery.BootstrapToken", "APIServerEndpoint", apiServerEndpoint)
+	}
+
+	// if BootstrapToken already contains a token, respect it; otherwise create a new bootstrap token for the node to join
+	if config.Spec.JoinConfiguration.Discovery.BootstrapToken.Token == "" {
+		// if there is no secretInterfaceFactory in place
+		if r.secretInterfaceFactory == nil {
+			// create the default secretInterfaceFactory function, that connects to the given remote cluster
+			// and then gets the secretInterface client for the kube-system namespace
+			r.secretInterfaceFactory = func(cluster *capiv1alpha2.Cluster) (corev1.SecretInterface, error) {
```
chuckha: This function is only closing around `r`. It takes cluster as a parameter and everything else is calculated. If you define this function on the `KubeadmConfigReconciler` it no longer has to be a function and this nil and anonymous function can go away.

If you're interested in unit testing this behavior I would recommend something like this

```go
type secretsClientFactory interface {
    secretsClient(*client.Client, *v1alpha2.Cluster)
}

type KubeadmConfigReconciler struct {
    ...
    secretsClientFactory secretsClientFactory
}
```

I'd be happy to discuss this on zoom if you'd like.

controllers/kubeadmconfig_controller.go
```diff
@@ -263,6 +267,83 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+		if err != nil {
+			return errors.Wrapf(err, "failed to create new bootstrap token")
+		}
+
+		config.Spec.JoinConfiguration.Discovery.BootstrapToken.Token = token
+		log.Info("Altering JoinConfiguration.Discovery.BootstrapToken", "Token", token)
+	}
+
+	// if BootstrapToken already contains a CACertHashes or UnsafeSkipCAVerification, respect it; otherwise set for UnsafeSkipCAVerification
+	// TODO: set CACertHashes instead of UnsafeSkipCAVerification
```
chuckha: Please file an issue for this as this must make it into the current milestone.

controllers/kubeadmconfig_controller_test.go
```diff
@@ -554,6 +561,180 @@ func TestReconcileIfJoinNodesAndControlPlaneIsReady(t *testing.T) {
...
+				},
+			},
+		},
+	}
+	var useCases = []struct {
+		name              string
+		cluster           *capiv1alpha2.Cluster
+		config            *cabpkV1alpha2.KubeadmConfig
+		validateDiscovery func(*testing.T, *cabpkV1alpha2.KubeadmConfig)
+		expectedError     bool
```
chuckha: I'm not picky on tests so this is non-blocking, but I wanted to offer some advice:

Make unit tests test exactly one thing. I have found the complexity of tests becomes unwieldy if this advice is not followed.

This one function could be at *least* two different `func Test..()` functions.

One of them could be `TestReconcileDiscoverSuccess` and `TestReconcilceDiscoverErrors`. This way your test logic has *no* conditionals.

controllers/kubeadmconfig_controller_test.go
```diff
@@ -554,6 +561,180 @@ func TestReconcileIfJoinNodesAndControlPlaneIsReady(t *testing.T) {
...
+
+			if err == nil && rt.expectedError {
+				t.Error("expected error, got nil")
+			}
+			if err != nil && !rt.expectedError {
+				t.Errorf("expected nil, got error %v", err)
+			}
+
+			if rt.validateDiscovery != nil {
+				rt.validateDiscovery(t, rt.config)
```
chuckha: Unless this function panics (it absolutely should not) there is no assertion here or the error is being ignored.

controllers/token.go
```diff
@@ -0,0 +1,68 @@
...
+// createToken attempts to create a token with the given ID.
+func createToken(client corev1.SecretInterface) (string, error) {
+	token, err := bootstraputil.GenerateBootstrapToken()
+	if err != nil {
+		return "", errors.Wrap(err, "unable to generate bootstrap token")
+	}
+
+	substrs := bootstraputil.BootstrapTokenRegexp.FindStringSubmatch(token)
+	if len(substrs) != 3 {
+		return "", errors.Wrapf(err, "the bootstrap token %q was not of the form %q", token, bootstrapapi.BootstrapTokenPattern)
```
chuckha: This should be `errors.Errorf` as there is no error to wrap, or rather, the err is nil anyway.

controllers/token.go
```diff
@@ -0,0 +1,68 @@
...
+			bootstrapapi.BootstrapTokenExpirationKey:       []byte(time.Now().UTC().Add(defaultTokenTTL).Format(time.RFC3339)),
+			bootstrapapi.BootstrapTokenUsageSigningKey:     []byte("true"),
+			bootstrapapi.BootstrapTokenUsageAuthentication: []byte("true"),
+			bootstrapapi.BootstrapTokenExtraGroupsKey:      []byte("system:bootstrappers:kubeadm:default-node-token"),
+			bootstrapapi.BootstrapTokenDescriptionKey:      []byte("token generated by cluster-api-bootstrap-provider-kubeadm"),
+		},
+	}
+
+	_, err = client.Create(secretToken)
+	return token, err
```
chuckha: If the secret fails to create is it still useful to return the generated token?

controllers/kubeadmconfig_controller.go
chuckha: Discussed offline. This is returning an error because it felt awkward to return (ctrl.Result{}, error). I suggested either returning (ctrl.Result{}, error) or custom errors and handling them in Reconcile.

controllers/kubeadmconfig_controller.go
chuckha: Discussed offline. We talked about where this should be initialized, I suggested the same point as the client for the reconciler as that's the only real dependency of this function.

controllers/kubeadmconfig_controller.go
fabriziopandini: There are previous art in cluster API of  nested reconcile funcs returning an error, and there is also an error type that specifically triggers requeue

controllers/kubeadmconfig_controller.go
fabriziopandini: aligned to the same format of controlPlane endpoint (AFAIK bootstrapToken apiserverendpoint accepts both)

controllers/kubeadmconfig_controller.go
fabriziopandini: Thanks for helping me in shaping this out!

controllers/kubeadmconfig_controller.go
fabriziopandini: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/123

controllers/token.go
```diff
@@ -0,0 +1,97 @@
...
+	capiremote "sigs.k8s.io/cluster-api/pkg/controller/remote"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	defaultTokenTTL = 10 * time.Minute
+)
+
+// NewClusterSecretsClientFactory returns a new ClusterSecretsClientFactory
+func NewClusterSecretsClientFactory() SecretsClientFactory {
```
chuckha: Why does this return a SecretsClientFactory? It looks like `ClusterSecretsClientFactory` satisfies the interface.

Best practice here would be to return concrete types and let the caller define the interface they need.

main.go
```diff
@@ -67,8 +67,9 @@ func main() {
 	}
 
 	if err := (&controllers.KubeadmConfigReconciler{
-		Client: mgr.GetClient(),
-		Log:    ctrl.Log.WithName("reconciler"),
+		Client:               mgr.GetClient(),
+		SecretsClientFactory: controllers.NewClusterSecretsClientFactory(),
```
chuckha: this should just be `controllers.ClusterSecretsClientFactory{}`

controllers/kubeadmconfig_controller_test.go
```diff
@@ -661,3 +868,18 @@ func newControlPlaneInitKubeadmConfig(machine *capiv1alpha2.Machine, name string
 func stringPtr(s string) *string {
 	return &s
 }
+
+func newFakeSecretFactory() SecretsClientFactory {
```
chuckha: This should return a concrete type.

The compiler will yell if FakeSecretFactory does not satisfy SecretsClientFactory.

fabriziopandini:
@chuckha last comment addressed + squash commits

PR #106, opened 2019-08-09
Use CAPI utils to reduce boilerplate

chuckha:

This PR uses the util code and removes some boilerplate making the Reconcile function shorter and easier to read.

It also bumps CAPI version and refactors some test logic.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/controllers/control_plane_init_locker_test.go
+++ b/controllers/control_plane_init_locker_test.go
@@ -33,6 +33,42 @@ import (
 
 func TestControlPlaneInitLockerAcquire(t *testing.T) {
 	tests := []struct {
+		name     string
+		getError error
+	}{
+		{
+			name:     "create succeeds",
+			getError: apierrors.NewNotFound(schema.GroupResource{Group: "", Resource: "configmaps"}, "uid1-configmap"),
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			l := &controlPlaneInitLocker{
+				log: log.ZapLogger(true),
+				configMapClient: &configMapsGetter{
+					getError: tc.getError,
+				},
+			}
+
+			cluster := &clusterv2.Cluster{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "ns1",
+					Name:      "name1",
+					UID:       types.UID("uid1"),
+				},
+			}
+
+			acquired := l.Acquire(cluster)
+			if !acquired {
+				t.Fatal("acquired was false but it should have been true")
+			}
+		})
+	}
+}
+
+func TestControlPlaneInitLockerAcquireErrors(t *testing.T) {
+	tests := []struct {
 		name          string
 		configMap     *v1.ConfigMap
 		getError      error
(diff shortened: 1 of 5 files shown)


controllers/kubeadmconfig_controller.go
```diff
@@ -76,29 +76,9 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
-	machineKey := client.ObjectKey{
-		Namespace: req.Namespace,
-		Name:      machineRef.Name,
-	}
-
-	if err := r.Get(ctx, machineKey, machine); err != nil {
-		log.Error(err, "failed to get machine")
+	machine, err := util.GetOwnerMachine(ctx, r.Client, config.ObjectMeta)
+	if err != nil {
+		log.Error(err, "could not get owner machine")
```
detiber: Should there be a check here for a `nil` machine? similar to the check included here: https://github.com/kubernetes-sigs/cluster-api-provider-docker/pull/173/files

chuckha: yes! I missed that one `return nil, nil`, good catch

controllers/kubeadmconfig_controller.go
```diff
@@ -76,49 +76,28 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (ctrl.Result, erro
...
-	clusterKey := client.ObjectKey{
-		Namespace: req.Namespace,
-		Name:      machine.Labels[capiv1alpha2.MachineClusterLabelName],
-	}
-	if err := r.Get(ctx, clusterKey, cluster); err != nil {
-		log.Error(err, "failed to get cluster")
+	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, machine.ObjectMeta)
+	if err != nil {
+		if err == util.ErrNoCluster {
+			log.Info(fmt.Sprintf("Please set the label %q on the machine equal to the name of the cluster this machine belongs to", capiv1alpha2.MachineClusterLabelName))
```
chuckha: is this the right place to log this message?

detiber: I would expect this error to be set from the `Machine` controller rather than here.

chuckha: ok, i'll nix it, thanks

controllers/control_plane_init_locker_test.go
```diff
@@ -32,6 +32,42 @@ import (
 )
 
 func TestControlPlaneInitLockerAcquire(t *testing.T) {
+	tests := []struct {
```
fabriziopandini: [1] Are you expecting more test case should be added? otherwise, we can get rid of subtests...

controllers/control_plane_init_locker_test.go
```diff
@@ -32,6 +32,42 @@ import (
...
+
+			acquired := l.Acquire(cluster)
+			if !acquired {
+				t.Fatal("acquired was false but it should have been true")
+			}
+		})
+	}
+}
+
+func TestControlPlaneInitLockerAcquireErrors(t *testing.T) {
```
fabriziopandini: [1]

chuckha: This one has 3 table tests, I don't think we should get rid of subtests.

fabriziopandini: time to take a break...

controllers/control_plane_init_locker_test.go
chuckha: some context:

I took a test that was testing too many things (error case + non error case) and split it into these two test functions. I kept the structure identical to make sure the tests didn't actually change.

I think it's ok to leave the table structure even with only one test.

My general approach here is to: Write a test that does not use table testing. If I need a second or third case I will add table tests if it makes sense, but overall I prefer the look and readability of table tests.

Are you ok leaving this one as is?

fabriziopandini: ok!

PR #110, opened 2019-08-10
chore: get kubeadm yaml rather than json

rudoi:
Currently, the userdata has the kubeadm config structs as json, which kubeadm doesn't seem to like.

Sample from a provisioned host:
```
---
{"nodeRegistration":{"kubeletExtraArgs":{"cni-bin-dir":"/opt/cni/bin","cni-conf-dir":"/etc/cni/net.d","network-plugin":"cni","node-ip":"<redacted>"}},"localAPIEndpoint":{"advertiseAddress":"","bindPort":0}}
```

To avoid importing k/k I grabbed a utility function and copied it here, MarshalToYamlForCodecs.

**Release note**:

```release-note

```

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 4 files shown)


chuckha:
the json should be fine🤔 do you have the error that you got handy? Is it related to the APIEndpoint change where `https://` is removed?

regardless yaml will be easier to debug for humans so I'm totally fine with this change.

detiber:
That should be able to be parsed as valid yaml, but like @chuckha I also think this change is warranted for readability.

rudoi:
@chuckha / @detiber wow uhhh y'all are right lol. I can't remember what error I hit but I can't seem to reproduce it. JSON works fine. Though I do agree the file looks nicer as YAML. Happy to change up anything if y'all want.

The `https://` thing is a real issue though, I believe.

PR #117, opened 2019-08-13
Add kubebuilder markers to CRDs for status and namespace

vincepri:

This PR adds kubebuilder directives to `KubeadmConfig` to ensure:
- Object is namespaced.
- Object has a `Status` subresource.
- The current version is marked as stored version.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/api/v1alpha2/kubeadmbootstrapconfig_types.go
+++ b/api/v1alpha2/kubeadmbootstrapconfig_types.go
@@ -49,6 +49,9 @@ type KubeadmConfigStatus struct {
 }
 
 // +kubebuilder:object:root=true
+// +kubebuilder:resource:path=kubeadmconfigs,scope=Namespaced
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
 
 // KubeadmConfig is the Schema for the kubeadmconfigs API
 type KubeadmConfig struct {
--- a/config/crd/bases/bootstrap.cluster.x-k8s.io_kubeadmconfigs.yaml
+++ b/config/crd/bases/bootstrap.cluster.x-k8s.io_kubeadmconfigs.yaml
@@ -10,7 +10,9 @@ spec:
   names:
     kind: KubeadmConfig
     plural: kubeadmconfigs
-  scope: ""
+  scope: Namespaced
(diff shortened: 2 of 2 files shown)


api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -49,6 +49,9 @@ type KubeadmConfigStatus struct {
 }
 
 // +kubebuilder:object:root=true
+// +kubebuilder:resource:path=kubeadmconfigs,scope=Namespaced
+// +kubebuilder:storageversion
```
chuckha: do we have conversion enabled?

what does this do when conversion is not enabled?

api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -49,6 +49,9 @@ type KubeadmConfigStatus struct {
 }
 
 // +kubebuilder:object:root=true
+// +kubebuilder:resource:path=kubeadmconfigs,scope=Namespaced
```
chuckha: `namespaced` is the default if I'm reading https://book.kubebuilder.io/reference/markers/crd.html correctly. I don't really know what this is doing otherwise. Why do we need it?

vincepri: It's actually not, it'll default to empty

api/v1alpha2/kubeadmbootstrapconfig_types.go
detiber: I think setting it proactively will help when we do enable conversions in the future

api/v1alpha2/kubeadmbootstrapconfig_types.go
chuckha: oh that's fun

api/v1alpha2/kubeadmbootstrapconfig_types.go
vincepri: The storageversion serves for conversion and for future version iterations. Other CAP* projects specify this field as well

PR #118, opened 2019-08-13
Streamline make generate

vincepri:

This PR aligns this Makefile with the other Cluster API repositories, where `make generate` runs a bunch of `*-generate` targets.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ all: manager
 
 # Run tests
 .PHONY: test
-test: generate fmt vet manifests
+test: generate fmt vet
 	go test ./api/... ./controllers/... -coverprofile cover.out
 
 # Build manager binary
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -91,8 +86,19 @@ vet:
...
+	$(MAKE) generate-manifests
+	$(MAKE) generate-deepcopy
+
+# Generate deepcopy files.
+.PHONY: generate-deepcopy
 	$(CONTROLLER_GEN) object:headerFile=./hack/boilerplate.go.txt paths=./api/...
 
+# Generate manifests e.g. CRD, RBAC etc.
+.PHONY: generate-manifests
+generate-manifests:
```
chuckha: This and generate-deepcopy need to depend on `$(CONTROLLER_GEN)` just like generate does. They should be able to run independently of `generate`.

Makefile
```diff
@@ -64,20 +64,15 @@ run: generate fmt vet
 
 # Install CRDs into a cluster
 .PHONY: install
-install: manifests
+install: generate
```
chuckha: this only depends on generate-manifests

Makefile
```diff
@@ -91,8 +86,19 @@ vet:
 # Generate code
 .PHONY: generate
 generate: $(CONTROLLER_GEN)
+	$(MAKE) generate-manifests
+	$(MAKE) generate-deepcopy
+
+# Generate deepcopy files.
+.PHONY: generate-deepcopy
```
chuckha: this is missing the actual target name

needs a `generate-deepcopy:` on the next line

PR #120, opened 2019-08-13
Update control-plane label value as cabpk-controller-manager

SataQiu:
Update `control-plane` label value as `cabpk-controller-manager`.
`controller-manager` can't tell what kind of controller manager it is, I think.

Fixes #

I found that the label `kube-controller-manager` uses is `component=kube-controller-manager`.
We can also consider adding a `component=cabpk-controller-manager` label.

**Release note**:

```release-note
NONE
```
/cc @chuckha @ncdc

### DIFF
--- a/config-capi/manager/manager.yaml
+++ b/config-capi/manager/manager.yaml
@@ -11,11 +11,11 @@ metadata:
   name: controller-manager-service
   namespace: system
   labels:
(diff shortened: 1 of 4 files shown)


chuckha:
nice, this has bugged me in the past.

/approve

PR #122, opened 2019-08-13
Only one node should do kubeadm init

### LINKED ISSUE #99: Prevent the bootstrapper to run kubeadm init on two nodes
/kind feature

As discussed on https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/92

while the controlplane is not ready, the controller should requeue the second bootstrap control plane and try to join it after the controlplane gets ready.

That means that if by chance the second bootstrap control plane has a JoinConfiguration, then it will join (InitConfiguration and Clusterconfiguration will be ignored at this point); If the second bootstrap control plane has not JoinConfiguration the controller will error out and eventually update ErrorMessage/ErrorReason.

The init-locker is already in the codebase, see https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/92; it is missing the plumbing into the controller

ykakarap:
This PR prevents the bootstrapper to run kubeadm init on two nodes

Fixes #99 

**Release note**:

```release-note

```

/assign @chuckha 
/assign @fabriziopandini

### DIFF
--- a/controllers/control_plane_init_locker.go
+++ b/controllers/control_plane_init_locker.go
@@ -17,14 +17,15 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/go-logr/logr"
 	apicorev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	clusterv2 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ControlPlaneInitLocker provides a locking mechanism for cluster initialization.
@@ -35,16 +36,18 @@ type ControlPlaneInitLocker interface {
 
 // controlPlaneInitLocker uses a ConfigMap to synchronize cluster initialization.
 type controlPlaneInitLocker struct {
-	log             logr.Logger
-	configMapClient corev1.ConfigMapsGetter
+	log    logr.Logger
+	client client.Client
+	ctx    context.Context
 }
 
 var _ ControlPlaneInitLocker = &controlPlaneInitLocker{}
 
-func newControlPlaneInitLocker(log logr.Logger, configMapClient corev1.ConfigMapsGetter) *controlPlaneInitLocker {
+func newControlPlaneInitLocker(ctx context.Context, log logr.Logger, client client.Client) *controlPlaneInitLocker {
 	return &controlPlaneInitLocker{
-		log:             log,
-		configMapClient: configMapClient,
+		ctx:    ctx,
+		log:    log,
+		client: client,
 	}
 }
 
@@ -77,7 +80,7 @@ func (l *controlPlaneInitLocker) Acquire(cluster *clusterv2.Cluster) bool {
 	}
 
 	log.Info("Attempting to create control plane configmap lock")
-	_, err = l.configMapClient.ConfigMaps(cluster.Namespace).Create(controlPlaneConfigMap)
+	err = l.client.Create(l.ctx, controlPlaneConfigMap)
 	if err != nil {
 		if apierrors.IsAlreadyExists(err) {
 			// Someone else beat us to it
@@ -99,7 +102,11 @@ func (l *controlPlaneInitLocker) Release(cluster *clusterv2.Cluster) bool {
 	log := l.log.WithValues("namespace", cluster.Namespace, "cluster-name", cluster.Name, "configmap-name", configMapName)
 
 	log.Info("Checking for existence of control plane configmap lock", "configmap-name", configMapName)
-	_, err := l.configMapClient.ConfigMaps(cluster.Namespace).Get(configMapName, metav1.GetOptions{})
+	cfg := &apicorev1.ConfigMap{}
+	err := l.client.Get(l.ctx, client.ObjectKey{
+		Namespace: cluster.Name,
+		Name:      configMapName,
+	}, cfg)
 	switch {
 	case apierrors.IsNotFound(err):
 		log.Info("Control plane configmap lock not found, it may have been released already", "configmap-name", configMapName)
@@ -107,7 +114,8 @@ func (l *controlPlaneInitLocker) Release(cluster *clusterv2.Cluster) bool {
 		log.Error(err, "Error retrieving control plane configmap lock", "configmap-name", configMapName)
 		return false
 	default:
-		if err := l.configMapClient.ConfigMaps(cluster.Namespace).Delete(configMapName, nil); err != nil {
+		// If no error delete the config map
+		if err := l.client.Delete(l.ctx, cfg); err != nil {
 			log.Error(err, "Error deleting control plane configmap lock", "configmap-name", configMapName)
 			return false
 		}
@@ -117,7 +125,12 @@ func (l *controlPlaneInitLocker) Release(cluster *clusterv2.Cluster) bool {
 }
 
 func (l *controlPlaneInitLocker) configMapExists(namespace, name string) (bool, error) {
-	_, err := l.configMapClient.ConfigMaps(namespace).Get(name, metav1.GetOptions{})
+	// _, err := l.configMapClient.ConfigMaps(namespace).Get(name, metav1.GetOptions{})
+	cfg := &apicorev1.ConfigMap{}
+	err := l.client.Get(l.ctx, client.ObjectKey{
+		Namespace: namespace,
+		Name:      name,
+	}, cfg)
 	if apierrors.IsNotFound(err) {
 		return false, nil
 	}
--- a/controllers/control_plane_init_locker.go
+++ b/controllers/control_plane_init_locker.go
@@ -125,7 +125,6 @@ func (l *controlPlaneInitLocker) Release(cluster *clusterv2.Cluster) bool {
 }
 
 func (l *controlPlaneInitLocker) configMapExists(namespace, name string) (bool, error) {
-	// _, err := l.configMapClient.ConfigMaps(namespace).Get(name, metav1.GetOptions{})
 	cfg := &apicorev1.ConfigMap{}
 	err := l.client.Get(l.ctx, client.ObjectKey{
 		Namespace: namespace,
--- a/controllers/control_plane_init_locker.go
+++ b/controllers/control_plane_init_locker.go
@@ -80,8 +80,7 @@ func (l *controlPlaneInitLocker) Acquire(cluster *clusterv2.Cluster) bool {
 	}
 
 	log.Info("Attempting to create control plane configmap lock")
-	err = l.client.Create(l.ctx, controlPlaneConfigMap)
-	if err != nil {
+	if err := l.client.Create(l.ctx, controlPlaneConfigMap); err != nil {
 		if apierrors.IsAlreadyExists(err) {
 			// Someone else beat us to it
 			log.Info("Control plane configmap lock already exists")
--- a/controllers/control_plane_init_locker_test.go
+++ b/controllers/control_plane_init_locker_test.go
@@ -17,219 +17,207 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
+	"errors"
+	"fmt"
 	"testing"
 
-	"github.com/pkg/errors"
-	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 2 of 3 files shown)


controllers/kubeadmconfig_controller.go
```diff
@@ -130,10 +133,13 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
-
-		// otherwise it is a init control plane
-		// Nb. in this case JoinConfiguration should not be defined by users, but in case of misconfigurations, CAPBK simply ignore it
+		// acquire the init lock so that only the first machine configured
+		// as control plane get processed here
+		// if not the first, requeue
+		if !initLocker.Acquire(cluster) {
+			log.Info("A control plane is already being initialized, requeing until control plane is ready")
+			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+		}
```
vincepri: You might want to `defer initLocker.Release(cluster)` here and remove the other calls, it should simplify a lot the code and avoid issues down the line :)

controllers/control_plane_init_locker.go
```diff
@@ -35,16 +36,18 @@ type ControlPlaneInitLocker interface {
 
 // controlPlaneInitLocker uses a ConfigMap to synchronize cluster initialization.
 type controlPlaneInitLocker struct {
-	log             logr.Logger
-	configMapClient corev1.ConfigMapsGetter
+	log    logr.Logger
+	client client.Client
+	ctx    context.Context
```
chuckha: Ah, you passed this into the struct because the `Acquire` interface does not allow for a context?

controllers/kubeadmconfig_controller.go
```diff
@@ -115,6 +115,9 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 		}
 	}()
 
+	// Creat the init locker
```
chuckha: this comment doesn't add much information, i'd be ok getting rid of it

controllers/control_plane_init_locker.go
```diff
@@ -117,7 +125,12 @@ func (l *controlPlaneInitLocker) Release(cluster *clusterv2.Cluster) bool {
 }
 
 func (l *controlPlaneInitLocker) configMapExists(namespace, name string) (bool, error) {
-	_, err := l.configMapClient.ConfigMaps(namespace).Get(name, metav1.GetOptions{})
+	// _, err := l.configMapClient.ConfigMaps(namespace).Get(name, metav1.GetOptions{})
```
chuckha: why is this commented out? I think deleting this would be preferred

controllers/kubeadmconfig_controller.go
chuckha: it looks like not *every* exit case needs to release the lock, just the errors?

controllers/control_plane_init_locker.go
ykakarap: will do.

controllers/kubeadmconfig_controller.go
```diff
@@ -152,6 +158,8 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 		initdata, err := json.Marshal(config.Spec.InitConfiguration)
 		if err != nil {
 			log.Error(err, "failed to marshal init configuration")
+			// release the lock
```
chuckha: I don't think we need these comments at every initLocker.Release call site. If you think `initLocker.Release` is not clear enough we should consider renaming it to something more clear

controllers/control_plane_init_locker.go
ykakarap: Yes. I didn't want to make changes to the original signature of `Acquire`.

controllers/kubeadmconfig_controller.go
ykakarap: will do

controllers/kubeadmconfig_controller.go
ykakarap: > it looks like not _every_ exit case needs to release the lock, just the errors?

You are correct Chuck. Only the error cases need to release the lock.

controllers/kubeadmconfig_controller.go
ykakarap: I think the function name is clear enough. I will remove the comments.

detiber:
@ykakarap you may want to incorporate some of the changes that @liztio is making here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1002

You wouldn't need the idempotencyToken, but the other aspects would likely be worth pulling in (especially around recording which resource the lock belongs to for reentrancy).

ykakarap:
Sure @detiber, I will check out the PR.

ykakarap:
> @ykakarap: The following test **failed**, say `/retest` to rerun them all:
> 
> Test name	Commit	Details	Rerun command
> pull-cluster-api-bootstrap-provider-kubeadm-verify	[7e49189](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/commit/7e49189e419b908186fa0e678f4235eb440d9e29)	[link](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-bootstrap-provider-kubeadm/122/pull-cluster-api-bootstrap-provider-kubeadm-verify/1161339220065783809)	`/test pull-cluster-api-bootstrap-provider-kubeadm-verify`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=cluster-api-bootstrap-provider-kubeadm&pr=122). [Your PR dashboard](https://gubernator.k8s.io/pr/ykakarap). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues?q=is:issue+is:open) when you hit one in your PR.

The tests are failing because the unit tests are broken because of the changes to how the init locker is created. Working on updating the tests.

ykakarap:
init locker unit tests are fixed and all tests are passing 🎉

controllers/kubeadmconfig_controller.go
neolit123: this could be:
```
releaseLock := true
...

if !initLocker.Acquire(cluster) {
	log.Info("A control plane is already being initialized, requeing until control plane is ready")
	releaseLock = false
	return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
}
...

defer func() {
	if releaseLock {
		someConditioninitLocker.Release(cluster)
	}
)
```

controllers/control_plane_init_locker_test.go
```diff
@@ -17,219 +17,207 @@ limitations under the License.
...
+			context:       context.Background(),
+			client:        &fakeClient{},
+			shouldAcquire: false,
 		},
 		{
-			name:          "create fails",
-			getError:      apierrors.NewNotFound(schema.GroupResource{Group: "", Resource: "configmaps"}, "uid1-configmap"),
-			createError:   errors.New("create error"),
-			expectAcquire: false,
+			name:    "shoult not acquire if cant create config map",
```
SataQiu:
```suggestion
			name:    "shoult not acquire lock if cannot create config map",
```

controllers/control_plane_init_locker_test.go
```diff
@@ -17,219 +17,207 @@ limitations under the License.
...
+			name:          "should release lock by deleting config map",
+			context:       context.Background(),
+			client:        &fakeClient{},
+			shouldRelease: true,
 		},
 		{
-			name:          "configmap not found",
-			getError:      apierrors.NewNotFound(schema.GroupResource{Group: "", Resource: "configmaps"}, "uid1-configmap"),
-			expectRelease: true,
+			name:    "should not release lock if cant delete config map",
```
SataQiu:
```suggestion
			name:    "should not release lock if cannot delete config map",
```

controllers/kubeadmconfig_controller.go
chuckha: @neolit123 the problem is still the same. Either we have to remember to release the lock at all the error return sites or remember to set releaseLock=true at all the non-error return sites.

controllers/control_plane_init_locker_test.go
```diff
@@ -17,219 +17,207 @@ limitations under the License.
...
 }
 
-func (c *configMapsGetter) ConfigMaps(namespace string) corev1client.ConfigMapInterface {
-	return &configMapClient{
-		configMap:   c.configMap,
-		getError:    c.getError,
-		createError: c.createError,
-		deleteError: c.deleteError,
-	}
+func (fsw *fakeStatusWriter) Patch(ctx context.Context, obj runtime.Object, patch client.Patch, opts ...client.PatchOption) error {
```
chuckha: the controller-runtime `client` package provides a fake client for us that we can use instead of this.

controllers/kubeadmconfig_controller.go
neolit123: my understanding is that the only branch that needs to set `releaseLock = false` is `if !initLocker.Acquire(cluster) {`

amy: This function is using named returns. Are you able to detect that there's an error and have defer release the lock based off that? 

```
defer func() {
  if rerr != nil {
    initLocker.Release(cluster)
  }
}
```

chuckha:
@ykakarap I think this needs a rebase. I'm ok with not using defer for now, but if you want to implement any one of the solutions that's fine too. Overall it looks good to me.

/approve

controllers/kubeadmconfig_controller.go
amy: Another option could be:
Pull that Locker logic into another function. There's a beginning and end to that locker logic anyways.

It could be something like this: 
```
func LockerLogic(...) err { // or whatever you want to name it
}

func Reconcile(...) {
  
  initLocker.Acquire(cluster)
  err := LockerLogic()
  initLocker.Release(cluster)
  if err != nil {
    return ctrl.Result{}, err
  }
```

}

amy:
Def out of scope for this PR. But this looks interesting in relation to your lock code: https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/resourcelock/leaselock.go

controllers/control_plane_init_locker_test.go
ykakarap: I looked at the fake client that the `client` package provides. I wanted to ability to throw errors in the `Create`, `Delete` and `Get` methods as needed so as to better unit test the locker. I don't think the fake client in the `client` package has this ability.

ykakarap:
@chuckha I rebased and also made some changes to put the lock release in a defer block. Please take a look.

controllers/control_plane_init_locker.go
```diff
@@ -35,16 +36,18 @@ type ControlPlaneInitLocker interface {
...
-	configMapClient corev1.ConfigMapsGetter
+	log    logr.Logger
+	client client.Client
+	ctx    context.Context
 }
 
 var _ ControlPlaneInitLocker = &controlPlaneInitLocker{}
 
-func newControlPlaneInitLocker(log logr.Logger, configMapClient corev1.ConfigMapsGetter) *controlPlaneInitLocker {
+func newControlPlaneInitLocker(ctx context.Context, log logr.Logger, client client.Client) *controlPlaneInitLocker {
```
vincepri: We should do this afterwards, but I think this code should move into its own package under `controllers/`, seems a little out of place.

controllers/control_plane_init_locker.go
```diff
@@ -77,7 +80,7 @@ func (l *controlPlaneInitLocker) Acquire(cluster *clusterv2.Cluster) bool {
 	}
 
 	log.Info("Attempting to create control plane configmap lock")
-	_, err = l.configMapClient.ConfigMaps(cluster.Namespace).Create(controlPlaneConfigMap)
+	err = l.client.Create(l.ctx, controlPlaneConfigMap)
 	if err != nil {
```
vincepri: Nit: prefer using `if err := <operation>; err != nil {` which scopes the variable `err` only to the following block

controllers/control_plane_init_locker_test.go
vincepri: I'd second using the `fake` client, and only interact with errors that come from using it, instead of dealing with arbitrary ones.

ykakarap: I explored using the `fake` client while writing these unit tests, however, some of the unit tests that need to be covered here are cases like "do not acquire the lock if cannot create config map". This means we need to replicate a case where the client's `Create` call fails. The `fake` client does not have a mechanism to set this up so as to test the case. There are similar cases in lock release as well. Hence I ended up creating this fake client in these unit tests.

We can use the `fake` client but that would mean we cannot test all the cases of the init locker.

controllers/control_plane_init_locker.go
ykakarap: I will move the init locker into a package under `controllers/` in a different PR.

ykakarap:
@vincepri I correct the nits you mentioned and addressed the `fake` client issue. Can you please re-review the PR? :)

ykakarap:
@chuckha @fabriziopandini This PR is ready.

PR #126, opened 2019-08-14
Add certificate generation and secret lookup to bootstrap controller

### LINKED ISSUE #74: Define CA pair secret names
/kind feature

In order to allow users to provide their own CA key pairs we should have this provider check a certain location for secrets. A user can upload secrets to this specific location and we will use them. If they don't exist we will create them and then use them.

This may pose a problem of certs disappearing between runs but I think it's ok to ignore that for right now.

Proposal:

| kind | secret name |
|---|---|
| Cluster CA Key | cluster-ca-key |
| Cluster CA Certificate | cluster-ca-cert |
| Etcd CA Key | etcd-ca-key |
| Etcd CA Certificate | etcd-ca-cert |
| FrontProxy CA Key | front-proxy-ca-key |
| FrontProxy CA Certificate | front-proxy-ca-cert |
| ServiceAccount Public Key | service-account-public-key |
| ServiceAccount Private Key | service-account-private-key |

The idea is to check for the secret, then generate if it's not there or empty and create it with data. If the secret is there and contains data then use that data blindly. We can consider making it smarter in the future.

jennydickinson:

Add certificate generation and secret lookup to bootstrap controller

Fixes #74 

**Release note**:

```release-note
Add certificate generation and secret lookup
```

@rudoi


vincepri:
Thanks for the contribution! 

What do you think about moving the `certs` package under `pkg/`? That way, we don't need to keep adding packages to the Dockerfile.

chuckha:
@vincepri let's have that discussion on an issue. it feels out of scope for this PR and i have some thoughts about why we should avoid pkg/ just like I'm sure you've got thoughts about why we should have pkg/

certs/files.go
```diff
@@ -14,94 +14,63 @@ See the License for the specific language governing permissions and
...
-	}
-
-	return nil
-}
+const (
+	rootOwnerValue = "root:root"
+)
 
-func certificatesToFiles(input Certificates) []v1alpha2.Files {
+func CertificatesToFiles(input Certificates) []v1alpha2.Files {
```
chuckha: Going to need to add godocs for every exported function

rudoi:
I've opened #127 to track the fact that we have two kinds of clients going. Most of this PR was already assembled when #105 was merged. If it's ok I'd like to address the consistency in a separate PR.

certs/utils.go
```diff
@@ -0,0 +1,78 @@
...
+		etcdCACertificate:        c.EtcdCA.Cert,
+		frontProxyCAKey:          c.FrontProxyCA.Key,
+		frontProxyCACertificate:  c.FrontProxyCA.Cert,
+		serviceAccountPrivateKey: c.ServiceAccount.Key,
+		serviceAccountPublicKey:  c.ServiceAccount.Cert,
+	}
+
+}
+
+func MapToCertificates(m map[string][]byte) *Certificates {
```
chuckha: I've generally seen this kind of constructor named as `NewCertificatesFromMap(`

certs/utils.go
```diff
@@ -0,0 +1,78 @@
...
+	clusterCACertificate     = "cluster-ca-cert"
+	etcdCAKey                = "etcd-ca-key"
+	etcdCACertificate        = "etcd-ca-cert"
+	frontProxyCAKey          = "front-proxy-ca-key"
+	frontProxyCACertificate  = "front-proxy-ca-cert"
+	serviceAccountPublicKey  = "service-account-public-key"
+	serviceAccountPrivateKey = "service-account-private-key"
+)
+
+func CertificatesToMap(c *Certificates) map[string][]byte {
```
chuckha: what's the reason behind making this a function instead of a method on `*Certficiates`?

controllers/kubeadmconfig_controller.go
```diff
@@ -330,27 +355,46 @@ func (r *KubeadmConfigReconciler) reconcileDiscovery(cluster *capiv1alpha2.Clust
...
-		EtcdCAKey:        string(certificates.EtcdCA.Key),
-		FrontProxyCACert: string(certificates.FrontProxyCA.Cert),
-		FrontProxyCAKey:  string(certificates.FrontProxyCA.Key),
-		SaCert:           string(certificates.ServiceAccount.Cert),
-		SaKey:            string(certificates.ServiceAccount.Key),
-	}, nil
+func (r *KubeadmConfigReconciler) getClusterCertificates(ctx context.Context, clusterName, namespace string) (*certs.Certificates, error) {
+	secret := &corev1.Secret{}
+
+	err := r.Get(ctx, types.NamespacedName{Name: fmt.Sprintf("%s-certs", clusterName), Namespace: namespace}, secret)
```
chuckha: what do you think about codifying this naming convention?

maybe move that fmt into a function with a name like `ClusterCertificatesSecretName` or something less verbose but still getting that point across? dunno if it's worth it if it's internal to this provider only tho.

controllers/kubeadmconfig_controller.go
```diff
@@ -330,27 +355,46 @@ func (r *KubeadmConfigReconciler) reconcileDiscovery(cluster *capiv1alpha2.Clust
...
+					Kind:       "KubeadmConfig",
+					Name:       config.GetName(),
+					UID:        config.GetUID(),
+				},
+			},
+		},
+		Data: certs.CertificatesToMap(certificates),
+	}
+
+	err = r.Create(ctx, secret)
```
chuckha: This should be one line like:

```
if err := Call(); err != nil {
}
```

This has the benefit of defining a new err var in the `if` scope which means we won't be reusing the err var. Reusing the err var makes it way too easy to accidentally not check every error return value

controllers/kubeadmconfig_controller.go
```diff
@@ -52,12 +55,13 @@ type KubeadmConfigReconciler struct {
...
 type SecretsClientFactory interface {
 	// NewSecretsClient returns a new client supporting SecretInterface
-	NewSecretsClient(client.Client, *capiv1alpha2.Cluster) (corev1.SecretInterface, error)
+	NewSecretsClient(client.Client, *capiv1alpha2.Cluster) (typedcorev1.SecretInterface, error)
 }
 
 // +kubebuilder:rbac:groups=bootstrap.cluster.x-k8s.io,resources=kubeadmconfigs,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=bootstrap.cluster.x-k8s.io,resources=kubeadmconfigs/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;machines,verbs=get;list;watch
+// +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch
```
chuckha: I think this change is only using get and create in which case this should be scoped down to the minimal necessary subset

controllers/kubeadmconfig_controller.go
```diff
@@ -330,27 +355,46 @@ func (r *KubeadmConfigReconciler) reconcileDiscovery(cluster *capiv1alpha2.Clust
...
+	certificates, err := certs.NewCertificates()
+	if err != nil {
+		return nil, err
+	}
+
+	secret := &corev1.Secret{
+		ObjectMeta: v1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-certs", clusterName),
+			Namespace: config.GetNamespace(),
+			OwnerReferences: []v1.OwnerReference{
```
chuckha: I think this means we'll need to start writing finalizers and cleaning this up on deletes now which is totally fine but I don't think we're doing that today? please open an issue for cleaning up properly

certs/files.go
chuckha: you did this while i was reviewing, resolving

controllers/kubeadmconfig_controller.go
```diff
@@ -330,27 +355,46 @@ func (r *KubeadmConfigReconciler) reconcileDiscovery(cluster *capiv1alpha2.Clust
...
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&cabpkv1alpha2.KubeadmConfig{}).
-		Complete(r)
+func (r *KubeadmConfigReconciler) createClusterCertificates(ctx context.Context, clusterName string, config *cabpkv1alpha2.KubeadmConfig) (*certs.Certificates, error) {
+	certificates, err := certs.NewCertificates()
+	if err != nil {
+		return nil, err
+	}
+
+	secret := &corev1.Secret{
```
moshloop: I think this will need to be 1 secret per cert to support cert-manager creating the certs: See `https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/74#issuecomment-518357317

controllers/kubeadmconfig_controller.go
detiber: There is a Create as well, so it's likely it will need all of them (except maybe patch)

controllers/kubeadmconfig_controller.go
detiber: Integration with cert-manager would likely need additional work and design, I'm not sure I would block this change on that.

controllers/kubeadmconfig_controller.go
chuckha: it is surprising to me that create use list, watch update and maybe patch, but that wouldn't be the first time a k8s client has surprised me 😂

controllers/kubeadmconfig_controller.go
detiber: I don't think we necessarily need make a finalizer here, the ownerReference should be good enough here for normal cascading deletion.

controllers/kubeadmconfig_controller.go
```diff
@@ -330,27 +355,46 @@ func (r *KubeadmConfigReconciler) reconcileDiscovery(cluster *capiv1alpha2.Clust
...
+		ObjectMeta: v1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-certs", clusterName),
+			Namespace: config.GetNamespace(),
+			OwnerReferences: []v1.OwnerReference{
+				{
+					// TODO: Consider whether this should have a cluster object as its owner
```
detiber: I think it is fine to just have the KubeadmConfig as an owner, since this is directly related to this provider.

controllers/kubeadmconfig_controller.go
chuckha: ah perfect! 👍

controllers/kubeadmconfig_controller.go
detiber: The controller runtime will use a cache, so list, watch, and possibly get are highly likely to be used.

As for patch, I don't necessarily *think* it will be used, but I'm not certain.

controllers/kubeadmconfig_controller.go
moshloop: If I run cert-manager (or anything else) and use it to pre-create the certs, and only then create the cluster it should work fine as long as the certificates are 1 per secret and named correctly. 

Integrating cert-manager with CAPI or this provider would require additional design, but that shouldn't stop people using it un-integrated

controllers/kubeadmconfig_controller.go
chuckha: i'm going to say this is resolved, it's fine leaving it as is 🎃

certs/utils.go
rudoi: This was my suggestion, only because I was annoyed by the fact that one direction would be a method and the other direction wouldn't be (without a new type) lol. 

I think this works just as well as `*Certificates.ToMap()`.

jennydickinson:
> nice!! this is awesome, thanks so much for this PR

It was a team effort, with @rudoi leading the charge. 😄 

I'll see what I can do about some of these fixes.

certs/utils.go
chuckha: sgtm

jennydickinson: I was already in the middle of changing this... might as well adhere to conventions.  Wouldn't want to stick out like a sore 👍

controllers/kubeadmconfig_controller.go
jennydickinson: How would you handle cases with more than one return value from `Call()`?

detiber: It depends...  if the other return values can be ignored or are only used within the if/else blocks, then I generally try to keep it on a single line. If that is not the case, then I would separate the call and test on different lines.

controllers/kubeadmconfig_controller.go
dennisme: Do we think that this is something that should block this PR? Or should we open an issue to make the secret format compatible with cert-manger?

detiber: +1 to creating an issue and handling any potential changes in post.

moshloop: This was agreed to as part of the issue, please see the comment history on the issue.

certs/files.go
chuckha: we may have interpreted that differently. I don't think we should block this PR on that change. We can update to use multiple secrets in a follow on so we can integrate with cert-manager and give us some time to design the solution.

controllers/kubeadmconfig_controller.go
jennydickinson: Ok... so it sounds like it's not _always_ possible to avoid reusing the err var.

certs/files.go
moshloop: If we can agree to block v1a2 on that new issue, then I am fine for this to merge. But if v1a2 gets released with secrets managed this way, then it creates upgrade and compatibility issues down the line which are completely avoidable.

chuckha: Ok, we'll work it out in the issue.

chuckha: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/133

controllers/kubeadmconfig_controller.go
dennisme: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/135 - happy to re word if its unclear or I missed something.

PR #128, opened 2019-08-14
AdditionalCommands support in kubeadm spec

### LINKED ISSUE #108: Add AdditionalCommands field to KubeadmConfigSpec
/kind feature

`ControlPlaneInput` and `NodeInput` both support `AdditionalCommands` in their teamplate and struct but it looks like `KubeadmConfigSpec` does not have a field that the user can use to take advantage of `AdditionalCommands`.

The `AdditionalCommands` field in `KubeadmConfigSpec` should eventually get wired up to `ControlPlaneInput` and `NodeInput`.

ykakarap:
This PR adds `AdditionalCommands` to Kubeadm spec and wires it up in the reconciler

Fixes #108 

**Release note**:

```release-note

```

/cc @chuckha

### DIFF
--- a/api/v1alpha2/kubeadmbootstrapconfig_types.go
+++ b/api/v1alpha2/kubeadmbootstrapconfig_types.go
@@ -33,9 +33,15 @@ type KubeadmConfigSpec struct {
 	// JoinConfiguration is the kubeadm configuration for the join command
 	// +optional
 	JoinConfiguration *kubeadmv1beta1.JoinConfiguration `json:"joinConfiguration,omitempty"`
-	// AdditionalUserDataFiles specifies extra files to be passed to user_data upon creation.
+	// Files specifies extra files to be passed to user_data upon creation.
 	// +optional
-	AdditionalUserDataFiles []Files `json:"additionalUserDataFiles,omitempty"`
+	Files []Files `json:"files,omitempty"`
+	// PreKubeadmCommands specifies extra commands to run before kubeadm runs
+	// +optional
+	PreKubeadmCommands []string `json:"preKubeadmCommands,omitempty"`
+	// PostKubeadmCommands specifies extra commands to run after kubeadm runs
+	// +optional
+	PostKubeadmCommands []string `json:"postKubeadmCommands,omitempty"`
 }
 
 // KubeadmConfigStatus defines the observed state of KubeadmConfig
--- a/cloudinit/cloudinit.go
+++ b/cloudinit/cloudinit.go
@@ -18,10 +18,10 @@ package cloudinit
 
 import (
 	"bytes"
-	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
 	"text/template"
 
 	"github.com/pkg/errors"
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
 )
 
 const (
@@ -32,10 +32,11 @@ const (
 
 // BaseUserData is shared across all the various types of files written to disk.
 type BaseUserData struct {
-	Header             string
-	AdditionalCommands []string
-	AdditionalFiles    []v1alpha2.Files
-	WriteFiles         []v1alpha2.Files
+	Header              string
+	PreKubeadmCommands  []string
+	PostKubeadmCommands []string
+	AdditionalFiles     []v1alpha2.Files
+	WriteFiles          []v1alpha2.Files
 }
 
 func generate(kind string, tpl string, data interface{}) ([]byte, error) {
--- a/api/v1alpha2/zz_generated.deepcopy.go
+++ b/api/v1alpha2/zz_generated.deepcopy.go
@@ -117,9 +117,19 @@ func (in *KubeadmConfigSpec) DeepCopyInto(out *KubeadmConfigSpec) {
 		*out = new(v1beta1.JoinConfiguration)
 		(*in).DeepCopyInto(*out)
 	}
-	if in.AdditionalUserDataFiles != nil {
-		in, out := &in.AdditionalUserDataFiles, &out.AdditionalUserDataFiles
+	if in.Files != nil {
+		in, out := &in.Files, &out.Files
 		*out = make([]Files, len(*in))
+		copy(*out, *in)
+	}
+	if in.PreKubeadmCommands != nil {
+		in, out := &in.PreKubeadmCommands, &out.PreKubeadmCommands
+		*out = make([]string, len(*in))
+		copy(*out, *in)
+	}
+	if in.PostKubeadmCommands != nil {
+		in, out := &in.PostKubeadmCommands, &out.PostKubeadmCommands
+		*out = make([]string, len(*in))
 		copy(*out, *in)
 	}
 }
--- a/cloudinit/controlplane_init.go
+++ b/cloudinit/controlplane_init.go
@@ -28,8 +28,9 @@ const (
       ---
 {{.InitConfiguration | Indent 6}}
 runcmd:
+{{- template "commands" .PreKubeadmCommands }}
   - 'kubeadm init --config /tmp/kubeadm.yaml'
-{{- template "commands" .AdditionalCommands }}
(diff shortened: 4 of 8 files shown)


api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -36,6 +36,9 @@ type KubeadmConfigSpec struct {
 	// AdditionalUserDataFiles specifies extra files to be passed to user_data upon creation.
 	// +optional
 	AdditionalUserDataFiles []Files `json:"additionalUserDataFiles,omitempty"`
+	// AdditionalCommands specifies extra commands to be passed to user_data upon creation
+	// +optional
+	AdditionalCommands []string `json:"additionalCommands,omitempty"`
```
detiber: I know previously we had discussed having both pre and post commands rather than just a single AdditionalCommands field. Could you update this for that? Maybe `AdditionalPreCommands` and `AdditionalPostCommands`?

Updating this file will also require updating the generated CRDs.

ykakarap: Supporting `AdditionalPreCommands`  and `AdditionalPostCommands` will also need changes to the cloud-init template, correct?

Something like this:
```
runcmd:
{{- template "commands" .AdditionalPreCommands }}
  - 'kubeadm init --config /tmp/kubeadm.yaml'
{{- template "commands" .AdditionalPostCommands }}
```

detiber: That would be correct, yes.

ykakarap: @detiber I was wondering if there an automatic way to generate the updated YAML using the updated type or maybe the reverse, generate the type using the YAML?

chuckha: if you modify the struct and run make generate the yaml should all get updated (not the cloud-init stuff though, that will be manual)

ykakarap:
@chuckha @detiber PR is ready for review

api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -36,6 +36,14 @@ type KubeadmConfigSpec struct {
 	// AdditionalUserDataFiles specifies extra files to be passed to user_data upon creation.
 	// +optional
 	AdditionalUserDataFiles []Files `json:"additionalUserDataFiles,omitempty"`
+	// AdditionalPreCommands specifies extra commands to be passed to user_data upon creation
+	// which run before kubeadm runs
+	// +optional
+	AdditionalPreCommands []string `json:"additionalPreCommands,omitempty"`
```
vincepri: What do we think about dropping the `Additional` prefix for these fields? Given that we have `Pre` and `Post` seems that should be fine?

cloudinit/cloudinit.go
```diff
@@ -18,9 +18,10 @@ package cloudinit
 
 import (
 	"bytes"
-	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
 	"text/template"
 
+	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
+
```
vincepri: Nit: Extra space

cloudinit/cloudinit.go
```diff
@@ -32,10 +33,11 @@ const (
 
 // BaseUserData is shared across all the various types of files written to disk.
 type BaseUserData struct {
-	Header             string
-	AdditionalCommands []string
-	AdditionalFiles    []v1alpha2.Files
-	WriteFiles         []v1alpha2.Files
+	Header                 string
+	AdditionalPreCommands  []string
+	AdditionalPostCommands []string
```
vincepri: Same here re:Additional prefix

api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -36,6 +36,14 @@ type KubeadmConfigSpec struct {
 	// AdditionalUserDataFiles specifies extra files to be passed to user_data upon creation.
 	// +optional
 	AdditionalUserDataFiles []Files `json:"additionalUserDataFiles,omitempty"`
+	// AdditionalPreCommands specifies extra commands to be passed to user_data upon creation
```
detiber: Not directly related to this PR, but we should likely avoid references to user_data, since that is an AWS specific term. @chuckha thoughts?

api/v1alpha2/kubeadmbootstrapconfig_types.go
chuckha: Interesting, i wonder why we had the word Additional commands in the first place? I guess to signify in addition to kubeadm init/join.

Does Additional add anything to this field? I think Additional adds a sense that these fields are optional. That if I don't need any additional commands I don't have to have them. But if they were just called Pre and Post I might feel more inclined like I need them.

Overall, I don't have any strong feelings and am happy with both Additional and Pre/Post. Up to @ykakarap if he'd like to update it or not.

api/v1alpha2/kubeadmbootstrapconfig_types.go
chuckha: Agreed, it should be `AdditionalFiles` I think, or perhaps simply `Files`. But yes, out of scope for this PR.

api/v1alpha2/kubeadmbootstrapconfig_types.go
ykakarap: Additional signifies that it runs in addition to kubeadm init/join command as mentioned by @chuckha .

If we want to drop "Additional" I suggest naming them `PreKubeadmCommands` and `PostKubeadmCommands`. This will suggest that they are commands besides the kubeadm commands and are also more descriptive of their behavior.

Thoughts?

chuckha: That also works. I was almost worried about tying the name to an implementation and then remembered this is the kubeadm bootstrap provider :)

I have no opinion on this. I'm happy to merge with Additional or Pre/Post KubeadmCommands

chuckha: @vincepri do you have strong feelings here?

fabriziopandini: I have not strong feeling, but If I have to express a preference is
- AdditionalFiles --> Files
- AdditionalPreCommands --> PreKubeadmCommands
- AdditionalPostCommands --> PostKubeadmCommands

vincepri: @fabriziopandini's suggestion looks good to me :)

api/v1alpha2/kubeadmbootstrapconfig_types.go
vincepri: I'd rephrase with something like 
```golang
... specifies extra commands to be run before kubeadm.
```

api/v1alpha2/kubeadmbootstrapconfig_types.go
ykakarap: >AdditionalFiles --> Files

`AdditionalFiles` is in `BaseUserData` struct. Shall i only change that to `Files` or also change the `AdditionalUserDataFiles` to `UserDataFiles` in `KubeadmConfigSpec` ?

chuckha: The important change is the KubeadmConfigSpec.

Ideally we don't reference UserData anywhere as that is a legacy AWS thing, but we can do that in a different PR and not change the API.

chuckha: To be explicit, please change `AdditionalUserDataFiles` to `Files` on the `KubeadmConfigSpec` type

ykakarap: Renamed the fields as discussed :)

ykakarap:
@chuckha @vincepri @fabriziopandini This PR is ready for review.

PR #130, opened 2019-08-14
Set the args in the manager_auth_proxy_patch.yaml

### LINKED ISSUE #129: Args cannot be set properly with kustomize
/kind bug

The manager_auth_proxy_patch sets args which overrides the args set in manager.yaml due to the strategic merge strategy.

One way to clean this up would be to only set args in the manager_auth_proxy_patch. Another might be to add `patchJson6902` to the kustomize.yaml and have a new patch that adds arguments to the manager image.

I expect to be able to set -v and --enable-leader-election on my manager.

/milestone v0.1
/priority important-soon

chuckha:

This PR is one solution to #129. It may not be optimal but does at least work.

Fixes #129

If this args is added to the manager_image_patch then there is a patch merge conflict.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -22,3 +22,5 @@ spec:
       - name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
+        - --v=4
+        - --enable-leader-election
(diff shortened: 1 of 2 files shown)


config/default/manager_auth_proxy_patch.yaml
```diff
@@ -22,3 +22,5 @@ spec:
       - name: manager
         args:
         - "--metrics-addr=127.0.0.1:8080"
+        - -v=5
```
SataQiu: SGTM. 
Found a nit.
```suggestion
        - --v=5
```

PR #134, opened 2019-08-15
Clean up some misleading logs

chuckha:

This PR cleans up some misleading logs and provides a nicer experience when a machine might be a control plane machine.

/assign @fabriziopandini @detiber 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -118,12 +118,11 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 		}
 	}()
 
-	// Check for control plane ready. If it's not ready then we will requeue the machine until it is.
-	// The cluster-api machine controller set this value.
+	// Init the control plane
 	if cluster.Annotations[ControlPlaneReadyAnnotationKey] != "true" {
 		// if it's NOT a control plane machine, requeue
 		if !util.IsControlPlaneMachine(machine) {
-			log.Info("Control plane is not ready, requeing worker nodes until ready.")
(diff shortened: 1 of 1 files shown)


controllers/kubeadmconfig_controller.go
```diff
@@ -121,12 +121,6 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 	// Check for control plane ready. If it's not ready then we will requeue the machine until it is.
 	// The cluster-api machine controller set this value.
 	if cluster.Annotations[ControlPlaneReadyAnnotationKey] != "true" {
-		// if it's NOT a control plane machine, requeue
-		if !util.IsControlPlaneMachine(machine) {
-			log.Info("Control plane is not ready, requeing worker nodes until ready.")
-			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
-		}
-
```
detiber: With removing this, should the log message on line 126 be updated as well to indicate that it will be requeueing both control plane and worker node machines?

chuckha: Ah ha, that log was copy pasta'd. Let me update that

chuckha: The comments were also not up-to-date so I took the liberty of updating those as well. The lock stuff is taken care of in #122

chuckha:
On second thought this PR is wrong.

My control plane was missing an annotation and the log line is actually incorrect which made me think this was unnecessary. Changing this PR, one minute.

chuckha:
done

detiber:
@chuckha description seems to have some copypasta as well:

> Add manager args to the proxoy patch

chuckha:
sigh. Sorry about that. Removed the extra sentence as well :/ doing too many things at once.

PR #146, opened 2019-08-19
Format field in KubeadmConfig.Spec

### LINKED ISSUE #121: Add a field to the spec called Format
/kind feature

Following up from a discussion https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/11#issuecomment-517698950

We are going to support multiple output versions *after* v1a2 but we are going to add the field now so that we don't have to modify the API later.

Don't forget to regenerate everything after adding the field!

/good-first-issue
/priority important-soon
/milestone v0.1

ykakarap:
This PR adds the field `Format` to `KubeadmConfig.Spec`. 
The `Format` is a required field that specifies the output format of the bootstrap data.

Fixes #121 

The `Format` field needs to be wired up. An issue will be opened to discuss how to use the field in the controller.

**Release note**:

```release-note

```

### DIFF
--- a/api/v1alpha2/kubeadmbootstrapconfig_types.go
+++ b/api/v1alpha2/kubeadmbootstrapconfig_types.go
@@ -21,6 +21,15 @@ import (
 	kubeadmv1beta1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
 )
 
+// Format specifies the output format of the bootstrap data
+// +kubebuilder:validation:Enum=cloud-config
+type Format string
+
+const (
+	// CloudConfig make the bootstrap data to be of cloud-config format
+	CloudConfig Format = "cloud-config"
+)
+
 // KubeadmConfigSpec defines the desired state of KubeadmConfig.
 // Either ClusterConfiguration and InitConfiguration should be defined or the JoinConfiguration should be defined.
 type KubeadmConfigSpec struct {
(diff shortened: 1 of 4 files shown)


api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -21,6 +21,12 @@ import (
 	kubeadmv1beta1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
 )
 
+type Format string
```
vincepri: We should add something like
`// +kubebuilder:validation:Enum=cloud-config`

ykakarap: Done :)

ykakarap:
I want to confirm that this field is not an optional field. It is a mandatory field.

detiber:
> For the sake of UX simplicity, IMO this flag should be optional/defaulted to cloud-config

Since we don't have the mutating webhook in place, we don't really have a good way to default today outside of the implementation. +1 to making this field optional, though.

api/v1alpha2/kubeadmbootstrapconfig_types.go
chuckha: nice

ykakarap:
@fabriziopandini @detiber I made the field optional.
As @detiber there is currently no mechanism to set a default. We have to make sure to set a default in the implementation while using the field.

detiber:
/lgtm
/hold
holding for further review by @fabriziopandini and @vincepri

chuckha:
lgtm pending the `make generate-manifests`

ykakarap:
>lgtm pending the make generate-manifests

I already did that. The YAML is already updated with the correct information. I wonder why it failed 🤔

Re-running tests.

/retest

chuckha:
@amy Any ideas on this test failure?

chuckha:
Discussed offline. This confusing error is because prow automatically rebases the patch.

chuckha:
hold for commit squashing

PR #151, opened 2019-08-20
add capi categories

mytunguyen:
Add capi categories so that we can do `kubectl get cluster-api`

Fixes #

**Release note**:

```release-note

```


go.mod
```diff
@@ -3,6 +3,7 @@ module sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm
 go 1.12
 
 require (
+	github.com/aws/aws-sdk-go v1.23.4 // indirect
```
chuckha: I think this might be leftover from working on CAPA?

mytunguyen: yup! that's not supposed to be there

PR #154, opened 2019-08-20
production ready dockerfile

fabriziopandini:
This PR renames the current `Dockerfile` into `Dockerfile.dev` and adds a production-ready  `Dockerfile` ("kubebuilder2" bases + GOPROXY + USER nobody) 

**Release note**:
```release-note
NONE
```

/assign @vincepri @chuckha

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 # Build the manager binary
-FROM golang:1.12.7 as builder
+FROM golang:1.12.7
 
 # default the go proxy
 ARG goproxy=https://proxy.golang.org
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -31,17 +32,19 @@ RUN go mod download
 # Copy the go source
 COPY main.go main.go
 COPY api/ api/
+COPY certs/ certs/
+COPY cloudinit/ cloudinit/
```
detiber: Why not just copy '.' instead?

vincepri: +1 to `COPY ./ ./`

fabriziopandini: I followed the kubebuilder template, but no strong preferences, so updated.

PR #155, opened 2019-08-20
add ErrorReason/ErrorStatus to CRD statuses

### LINKED ISSUE #97: Add ErrorReason/ErrorStatus to CRD Statuses
/kind feature

our configs can end up in an error state, such as having a cluster with two configs that both only have join & init configuration. In this case we need to be able to report up that one is broken and needs attention.

In order to do this we will need ErrorReason and ErrorStatus on our CRDs.

Look here but do not use the same capierrors.MachineStatusError type since that doesn't apply to this case.

https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha2/machine_types.go#L111

Also please regenerate CRDs after the new fields are added.

/priority important-longterm
/milestone v0.1
^ note on milestone, it can slip due to lower priority

mytunguyen:
Add ErrorReason/ErrorStatus to CRD statuses so it's easier to debug. 

Fixes #97 

**Release note**:

```release-note

```

### DIFF
--- a/api/v1alpha2/kubeadmbootstrapconfig_types.go
+++ b/api/v1alpha2/kubeadmbootstrapconfig_types.go
@@ -64,6 +64,14 @@ type KubeadmConfigStatus struct {
 	// BootstrapData will be a cloud-init script for now
 	// +optional
 	BootstrapData []byte `json:"bootstrapData,omitempty"`
+
+	// ErrorReason will be set on non-retryable errors
+	// +optional
+	ErrorReason string `json:"errorReason,omitempty"`
+
+	// ErrorMessage will be set on non-retryable errors
+	// +optional
+	ErrorMessage string `json:"errorMessage,omitempty"`
 }
 
 // +kubebuilder:object:root=true
--- a/config/crd/bases/bootstrap.cluster.x-k8s.io_kubeadmconfigs.yaml
+++ b/config/crd/bases/bootstrap.cluster.x-k8s.io_kubeadmconfigs.yaml
@@ -728,6 +728,12 @@ spec:
               description: BootstrapData will be a cloud-init script for now
               format: byte
               type: string
+            errorMessage:
+              description: ErrorMessage will be set on non-retryable errors
+              type: string
+            errorReason:
+              description: ErrorReason will be set on non-retryable errors
+              type: string
             ready:
               description: Ready indicates the BootstrapData field is ready to be
                 consumed


vincepri:
LGTM

/ok-to-test

controllers/kubeadmconfig_controller.go
```diff
@@ -95,6 +96,8 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 	machine, err := util.GetOwnerMachine(ctx, r.Client, config.ObjectMeta)
 	if err != nil {
 		log.Error(err, "could not get owner machine")
+                config.Status.ErrorStatus = CreateKubeadmconfigError
```
detiber: Should this be a non-retryable error? I would probably expect this to be a transient error only.

We should only set ErrorMessage and ErrorReason for non-retryable errors.

vincepri: Agreed, I missed this change when I first glanced through the PR.

@mytunguyen ErrorStatus and ErrorReason are usually only set in failures, when you usually require user intervention for example.

controllers/kubeadmconfig_controller.go
```diff
@@ -95,6 +96,8 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+                config.Status.ErrorReason = "could not get owner machine"
 		return ctrl.Result{}, err
```
chuckha: should this even be returning err?

ncdc: No, I don't think it should. We should wait for the CAPI machine controller to set the ownerReference in the KubeadmConfig to the Machine. Once that happens, this controller will get a natural update event and rereconcile the KubeadmConfig.

detiber: Agreed, no need to return an error here.

controllers/kubeadmconfig_controller.go
mytunguyen: makes sense. okay!

controllers/kubeadmconfig_controller.go
mytunguyen: all set I think

vincepri:
/assign @chuckha 

for approval

chuckha:
@vincepri What do you think about aligning this with the Machine and Cluster type from CAPI and using `ErrorReason` && `ErrorMessage`? I'm still trying to figure out if `ErrorReason` even makes sense here 🤔  (the machine readable error message. what would actually be consuming this error message? i think this is only for humans?)

vincepri:
+1 for consistency, the plan sounds good to me. I'd keep both for now, until we actually have a large failure-mode handling plan

chuckha:
/approve

Thanks for adding these!

chuckha:
oh, can you squash your commits?

mytunguyen:
Sweet! Did it

vincepri:
Good timing :D

PR #161, opened 2019-08-21
Add ntp and user module support for cloud-init

### LINKED ISSUE #100: Ability to configure users via cloud-config
/kind feature

I would like to be able to define information about users as a part of the cloud-config data.

For example, I may want to add a service account with a specific username, SSH authorized key, and group so that I can use to connect to nodes after they have been provisioned.

For example, I may want to restrict the root user's configuration.

A golang template for configuring this could look something like:
```
{{- if .Users }}
users:
  {{- range .Users }}
  - name: {{ .Name }}
    shell: {{ .Shell  }}
    passwd: {{ .HashedPassword }}
    lock_passwd: {{ .LockPassword }}
    {{- if .Sudo }}
    sudo: {{ .Sudo }}
    {{- end }}
    {{- if .SshAuthorizedKeys }}
    ssh_authorized_keys:
    {{- range .SshAuthorizedKeys }}
      - {{ . }}
    {{- end }}
    {{- end }}
  {{- end }}
{{- end }}
```

### LINKED ISSUE #101: Ability to configure NTP servers via cloud-config
/kind feature

I would like to be able to define NTP servers as a part of cloud-config.

I may have NTP servers internal to my datacenter which I want my nodes to use.

A golang template for configuring this could look something like:

```
{{- if .NTPServers }}
ntp:
  enabled: true
  servers:
  {{- range .NTPServers }}
  - {{ . }}
  {{- end }}
{{- end }}
```

detiber:
Adds support for ntp and user modules when generating cloud-init configuration

Fixes #100, Fixes #101

**Release note**:
```release-note
KubeadmConfig now supports specifying ntp configuration and additional users when generating cloud-init configuration.
```


api/v1alpha2/kubeadmbootstrapconfig_types.go
```diff
@@ -51,6 +51,12 @@ type KubeadmConfigSpec struct {
 	// PostKubeadmCommands specifies extra commands to run after kubeadm runs
 	// +optional
 	PostKubeadmCommands []string `json:"postKubeadmCommands,omitempty"`
+	// Users specifies extra users to add
+	// +optional
+	Users []User `json:"users,omitempty"`
```
chuckha: Are we tying or have we already tied this struct too tightly to cloud-init? It might be good if we had a cloud-init substructure to host the cloud-init specific bits. Stuff bash wouldn't care about.

detiber: I don't think we want to get into supporting different subsets of configuration based on output format, otherwise it'll be a mess for users.

chuckha: ok. i'm not terribly picky so long as they are all optional it will still look ok to a user

chuckha:
/lgtm

🚀

PR #188, opened 2019-08-29
🐛 Fix a Cluster -> KubeadmConfig mapper bug

vincepri:

Fixes #

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -44,14 +44,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
-// KubeadmConfigReconciler reconciles a KubeadmConfig object
-type KubeadmConfigReconciler struct {
-	client.Client
-	SecretsClientFactory SecretsClientFactory
-	KubeadmInitLock      InitLocker
-	Log                  logr.Logger
-}
-
 // InitLocker is a lock that is used around kubeadm init
 type InitLocker interface {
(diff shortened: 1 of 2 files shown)


controllers/kubeadmconfig_controller.go
```diff
@@ -378,15 +377,23 @@ func (r *KubeadmConfigReconciler) ClusterToKubeadmConfigs(o handler.MapObject) [
...
 		r.Log.Error(err, "failed to list Machines", "Cluster", c.Name, "Namespace", c.Namespace)
 		return nil
 	}
+
 	for _, m := range machineList.Items {
-		if m.Spec.InfrastructureRef.GroupVersionKind() == bootstrapv1.GroupVersion.WithKind("KubeadmConfig") {
-			name := client.ObjectKey{Namespace: m.Namespace, Name: m.Spec.InfrastructureRef.Name}
+		if m.Spec.Bootstrap.ConfigRef != nil &&
+			m.Spec.Bootstrap.ConfigRef.GroupVersionKind() == bootstrapv1.GroupVersion.WithKind("KubeadmConfig") {
+			name := client.ObjectKey{Namespace: m.Namespace, Name: m.Spec.Bootstrap.ConfigRef.Name}
```
detiber: This diff is hard to read with all the additional refactoring as well as the bugfix, but afaict, this is the heart of the fix here where we were previously mistakenly queuing the InfrastructureRef.Name rather than the ConfigRef.Name. Is that correct?

vincepri: yeah

PR #211, opened 2019-09-05
🐛 Test and logic fixes

chuckha:

Some more test clean ups and the discovery of a bug where the user could not set UnsafeSkipCAVerification even if they wanted to. In addition, this adds error handling for a missed error check.


controllers/kubeadmconfig_controller.go
```diff
@@ -267,10 +267,15 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
 		hashes, err := certs.CertificateHashes(certificates.ClusterCA.Cert)
-		if err == nil {
-			config.Spec.JoinConfiguration.Discovery.BootstrapToken = &kubeadmv1beta1.BootstrapTokenDiscovery{
-				CACertHashes: hashes,
+		if err != nil {
+			log.Error(err, "Unable to generate certificate hashes")
+			return ctrl.Result{}, err
+		}
+		if hashes != nil {
+			if config.Spec.JoinConfiguration.Discovery.BootstrapToken == nil {
```
fabriziopandini: IMO this piece of code should be moved into `reconcileDiscovery `, so all the logics impaction `JoinConfiguration.Discovery` are in the same place.
Nb. the bit of code for creating `BootstrapToken` if nil is already implemented there as well

chuckha: I don't think it makes sense to add this check to `reconcileDiscovery` unless you are also suggesting to add certificates into the `reconcileDiscovery` method. We look up certificates to see if we have a CACertHash to use. If we do then `reconcileDiscovery` does something different than if it didn't have a CACertHash. Therefore `reconcileDiscovery` depends on looking up certificates. Either certificates becomes part of `reconcileDiscovery` or we will have to live with not all defaults ending up in `reconcileDiscovery`

If we want to refactor it that's fine but I think that's out of scope for this PR.

The order is a little bit weird because we call `reconcileDiscovery right after we figure out our Certificates

controllers/kubeadmconfig_controller_test.go
```diff
@@ -374,7 +377,7 @@ func TestKubeadmConfigReconciler_Reconcile_ErrorIfAWorkerHasNoJoinConfigurationA
 }
 
 // If a controlplane has an invalid JoinConfiguration then user intervention is required.
-// TODO: Could potentially requeue instead of error.
+// TODO: Could potentially requeue/return nothing instead of error.
```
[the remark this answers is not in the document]
chuckha: Agreed. I was thinking about technically it will requeue when the user fixes it...but if the user doesn't know it's broken they won't fix it 😆

controllers/kubeadmconfig_controller_test.go
```diff
@@ -339,6 +340,8 @@ func TestKubeadmConfigReconciler_Reconcile_GenerateCloudConfigData(t *testing.T)
 	}
 }
 
+// Return an error if a worker has no JoinConfiguration defined
+// TODO: This logic should probably just return and wait for the worker to have a join configuration?
```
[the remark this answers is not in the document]
chuckha: Agreed, updating comment.

controllers/kubeadmconfig_controller.go
```diff
@@ -432,9 +437,8 @@ func (r *KubeadmConfigReconciler) reconcileDiscovery(cluster *clusterv1.Cluster,
 		log.Info("Altering JoinConfiguration.Discovery.BootstrapToken", "Token", token)
 	}
 
-	// if BootstrapToken already contains a CACertHashes or UnsafeSkipCAVerification, respect it; otherwise set for UnsafeSkipCAVerification
-	// TODO: set CACertHashes instead of UnsafeSkipCAVerification
-	if len(config.Spec.JoinConfiguration.Discovery.BootstrapToken.CACertHashes) == 0 && !config.Spec.JoinConfiguration.Discovery.BootstrapToken.UnsafeSkipCAVerification {
+	// if BootstrapToken does not contain any CACertHashes then force skip CA Verification
```
vincepri: Smallest of nits 😬 
```suggestion
	// If BootstrapToken does not contain any CACertHashes then force skip CA Verification
```

chuckha: Updated

controllers/kubeadmconfig_controller_test.go
```diff
@@ -858,34 +867,66 @@ func TestCACertHashesAndUnsafeCAVerifySkip(t *testing.T) {
...
+		},
+		{
+			name: "Skip CA verification if requested by the user",
+			discovery: &kubeadmv1beta1.BootstrapTokenDiscovery{
+				UnsafeSkipCAVerification: true,
+			},
+			skipCAVerification: true,
+		},
+		{
+			// This should be true since no Cert Hashes are provided, but reconcile will *always* get or create certs.
```
liztio: what should be "true" here?

controllers/kubeadmconfig_controller_test.go
```diff
@@ -815,39 +818,45 @@ func TestReconcileTopLevelObjectSettings(t *testing.T) {
...
-		t.Run(rt.name, func(t *testing.T) {
-			k.reconcileTopLevelObjectSettings(rt.cluster, rt.machine, rt.config)
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			k.reconcileTopLevelObjectSettings(tc.cluster, tc.machine, tc.config)
 
-			if rt.config.Spec.ClusterConfiguration.ControlPlaneEndpoint != "myControlPlaneEndpoint:6443" {
-				t.Fatalf("expected ClusterConfiguration.ControlPlaneEndpoint %q, got %q", "myControlPlaneEndpoint:6443", rt.config.Spec.ClusterConfiguration.ControlPlaneEndpoint)
+			if tc.config.Spec.ClusterConfiguration.ControlPlaneEndpoint != "myControlPlaneEndpoint:6443" {
+				t.Fatalf("expected ClusterConfiguration.ControlPlaneEndpoint %q, got %q", "myControlPlaneEndpoint:6443", tc.config.Spec.ClusterConfiguration.ControlPlaneEndpoint)
```
liztio: These could all be Errorf pretty safely, I think

controllers/kubeadmconfig_controller_test.go
chuckha: fair. Fixed.

controllers/kubeadmconfig_controller_test.go
chuckha: for sure, updated.

PR #219, opened 2019-09-10
🏃 Move certs to an internal package

chuckha:

This PR moves certs to the internal package because we do not want anyone consuming this package as a library, nor do we have any capacity to provide support for this as a library.


detiber:
@chuckha I believe this is conflicts with https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1081

chuckha:
dear oh dear. Ok, never mind on this one then...

/close

Issue #214, opened 2019-09-05
Testing for expected behaviors


chuckha:
/kind feature

**Describe the solution you'd like**
We've done a decent job of testing very important behaviors but I'd like to add some more tests for a few more behaviors listed here:

⚠️ In order to view the [original document](https://docs.google.com/document/d/1g6xxhD_AFA4klrm87f7o0XpEtOOZfbSIat3ltKmbvrc/edit
) please join the [kubernetes-dev](https://groups.google.com/forum/#!forum/kubernetes-dev) or the [kubernetes-sig-cluster-lifecycle](https://groups.google.com/forum/#!forum/kubernetes-sig-cluster-lifecycle) google groups. ⚠️

The tests I'd like to see added are:

Note, you may be able to argue some of these exist or I may have missed some scenarios and these don't need implementing. That is ok too!

## Tests

- [x] Return early if the owning machine does not have an associated cluster (@hpandeycodeit https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/240)
- [x] Return early if the cluster’s infrastructure is not ready (@sb1975)
- [x] Exactly one control plane machine initializes if there are multiple control plane machines defined (there are tests for the lock but not explicit tests here) (@SataQiu https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/221)
- [x] Uses existing certificate secrets (@Xenwar https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/235)
- [ ] Generates a valid cloud-config (**TODO**)
  - [ ] Include characters that require escaping (**TODO**)
- [x] Triggers a reconcile for an associated KubeadmConfig when a Machine that references it is modified (@mikkosest #237 )
- [x] Triggers a reconcile for all KubeadmConfigs associated with all Machines associated with a modified Cluster. (@chuckha #216)

hpandeycodeit:
I will work on `Return early if the owning machine does not have an associated cluster ` :)

Xenwar:
Do I take 'Uses existing certificate secrets' ?
thanks.

chuckha:
@Xenwar Sure thing! let me know if you have any questions or need help, I'm `@cha` on the k8s slack.

sb1975:
Reg this Test :  Return early if the cluster’s infrastructure is not ready (TODO)
[...]

PR #221, opened 2019-09-11
🏃‍♂️ Add a test to ensure that exactly one control plane machine initializes if there are multiple control plane machines defined

SataQiu:
Add a test to ensure that exactly one control plane machine initializes if there are multiple control plane machines defined

Ref #214

### DIFF
--- a/controllers/kubeadmconfig_controller_test.go
+++ b/controllers/kubeadmconfig_controller_test.go
@@ -119,13 +119,13 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfKubeadmConfigIsReady(t *
 	}
 	result, err := k.Reconcile(request)
 	if err != nil {
-		t.Fatal(fmt.Sprintf("Failed to reconcile:\n %+v", err))
+		t.Fatalf("Failed to reconcile:\n %+v", err)
 	}
 	if result.Requeue == true {
-		t.Fatal("did not expected to requeue")
+		t.Fatal("did not expect to requeue")
 	}
 	if result.RequeueAfter != time.Duration(0) {
-		t.Fatal("did not expected to requeue after")
+		t.Fatal("did not expect to requeue after")
 	}
 }
 
@@ -182,13 +182,13 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasBootstrapData(
 	}
 	result, err := k.Reconcile(request)
 	if err != nil {
-		t.Fatal(fmt.Sprintf("Failed to reconcile:\n %+v", err))
+		t.Fatalf("Failed to reconcile:\n %+v", err)
 	}
 	if result.Requeue == true {
-		t.Fatal("did not expected to requeue")
+		t.Fatal("did not expect to requeue")
 	}
 	if result.RequeueAfter != time.Duration(0) {
-		t.Fatal("did not expected to requeue after")
+		t.Fatal("did not expect to requeue after")
 	}
 }
 
@@ -258,8 +258,8 @@ func TestKubeadmConfigReconciler_Reconcile_RequeueJoiningNodesIfControlPlaneNotI
 	workerMachine := newWorkerMachine(cluster)
 	workerJoinConfig := newWorkerJoinKubeadmConfig(workerMachine)
 
-	controlPlaneMachine := newControlPlaneMachine(cluster)
-	controlPlaneJoinConfig := newControlPlaneJoinKubeadmConfig(controlPlaneMachine, "control-plane-join-cfg")
+	controlPlaneJoinMachine := newControlPlaneMachine(cluster, "control-plane-join-machine")
+	controlPlaneJoinConfig := newControlPlaneJoinKubeadmConfig(controlPlaneJoinMachine, "control-plane-join-cfg")
 
 	testcases := []struct {
 		name    string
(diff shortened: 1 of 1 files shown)


controllers/kubeadmconfig_controller_test.go
```diff
@@ -970,6 +970,67 @@ func TestKubeadmConfigReconciler_ClusterToKubeadmConfigs(t *testing.T) {
 	}
 }
 
+// Exactly one control plane machine initializes if there are multiple control plane machines defined
+func TestKubeadmConfigReconciler_Reconcile_ExactlyOneControlPlaneMachineInitializes(t *testing.T) {
```
neolit123: side topic: i think that `_` in function names fails `golint`
is this file ignored from golint checks?

controllers/kubeadmconfig_controller_test.go
```diff
@@ -970,6 +970,67 @@ func TestKubeadmConfigReconciler_ClusterToKubeadmConfigs(t *testing.T) {
...
+
+	request := ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: "default",
+			Name:      "control-plane-init-cfg-first",
+		},
+	}
+	result, err := k.Reconcile(request)
+	if err != nil {
+		t.Fatal(fmt.Sprintf("Failed to reconcile:\n %+v", err))
```
neolit123: use `Fatalf` instead of `Fatal + Sprintf`

controllers/kubeadmconfig_controller_test.go
```diff
@@ -970,6 +970,67 @@ func TestKubeadmConfigReconciler_ClusterToKubeadmConfigs(t *testing.T) {
...
+			Namespace: "default",
+			Name:      "control-plane-init-cfg-second",
+		},
+	}
+	result, err = k.Reconcile(request)
+	if err != nil {
+		t.Fatal(fmt.Sprintf("Failed to reconcile:\n %+v", err))
+	}
+	if result.Requeue == true {
+		t.Fatal("did not expected to requeue")
```
neolit123: did not expect

controllers/kubeadmconfig_controller_test.go
```diff
@@ -970,6 +970,67 @@ func TestKubeadmConfigReconciler_ClusterToKubeadmConfigs(t *testing.T) {
...
+
+	request = ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: "default",
+			Name:      "control-plane-init-cfg-second",
+		},
+	}
+	result, err = k.Reconcile(request)
+	if err != nil {
+		t.Fatal(fmt.Sprintf("Failed to reconcile:\n %+v", err))
```
neolit123: [1]

controllers/kubeadmconfig_controller_test.go
```diff
@@ -970,6 +970,67 @@ func TestKubeadmConfigReconciler_ClusterToKubeadmConfigs(t *testing.T) {
...
+	}
+	result, err := k.Reconcile(request)
+	if err != nil {
+		t.Fatal(fmt.Sprintf("Failed to reconcile:\n %+v", err))
+	}
+	if result.Requeue == true {
+		t.Fatal("did not expected to requeue")
+	}
+	if result.RequeueAfter != time.Duration(0) {
+		t.Fatal("did not expected to requeue after")
```
neolit123: did not expect

controllers/kubeadmconfig_controller_test.go
```diff
@@ -970,6 +970,67 @@ func TestKubeadmConfigReconciler_ClusterToKubeadmConfigs(t *testing.T) {
...
+			Namespace: "default",
+			Name:      "control-plane-init-cfg-first",
+		},
+	}
+	result, err := k.Reconcile(request)
+	if err != nil {
+		t.Fatal(fmt.Sprintf("Failed to reconcile:\n %+v", err))
+	}
+	if result.Requeue == true {
+		t.Fatal("did not expected to requeue")
```
neolit123: did not expect

controllers/kubeadmconfig_controller_test.go
```diff
@@ -1106,9 +1167,21 @@ func (f FakeSecretFactory) NewSecretsClient(client client.Client, cluster *clust
...
+type myInitLocker struct {
+	locked bool
+}
 
 func (m *myInitLocker) Lock(_ context.Context, _ *clusterv1.Cluster, _ *clusterv1.Machine) bool {
+	if !m.locked {
+		m.locked = true
+		return true
+	}
+	return false
```
neolit123: should we always return `true` for the fake Lock?

controllers/kubeadmconfig_controller_test.go
chuckha: I'm not sure actually..i didn't see anything in the code that indicated tests are skipped. I'd say let's not worry about it for this PR. I've made an issue for it https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/223

chuckha:
This looks great to me pending @neolit123's changes.

The fake lock is a little bit weird, would it be possible to always return true? a simpler implementation would be better, but if this is necessary for now then let's leave it.

SataQiu:
> 
> 
> This looks great to me pending @neolit123's changes.
> 
> The fake lock is a little bit weird, would it be possible to always return true? a simpler implementation would be better, but if this is necessary for now then let's leave it.

Thanks @chuckha 
I don't think the fake lock should always return true. 
Because only when the lock is unlocked can the initializing machine obtain the lock.

chuckha:
/approve

thanks for the update this looks really good!

/assign @neolit123 
for lgtm

SataQiu:
ping @neolit123

SataQiu:
Hi @ncdc 
Could you have a look at this? 
This PR has been around for a long time

PR #222, opened 2019-09-11
⚠️ Certificate extraction

### LINKED ISSUE #59: Improve user experience when using an external etcd
We must write a few additional certificate files.

The user will be supplying the `etcd.crt` file in addition to the `apiserver-etcd-client.key` and `apiserver-etcd-client.crt` files. They will use the normal etcd secret and CABPK will look for an additional secret, `<cluster-name>-etcd-client` (may not be this name exactly). It will then write these secrets to disk depending on the configuration of the externalEtcd object on the ClusterConfiguration option.

<details>
<summary><b>Original issue click me!</b></summary>
/kind feature

The current implementation plans for HA consist of deploying using a stacked etcd control plane using the `kubeadm join --experimental-controlplane` feature. There is also a need to eventually support HA clusters where etcd is hosted separately from the control plane.

To simplify the maintenance burden of this approach, we should rely on [etcdmgr](https://github.com/kubernetes/community/blob/master/keps/sig-cluster-lifecycle/0031-20181022-etcdadm.md) for managing the etcd components rather than rolling our own custom etcd management.
</details>

### LINKED ISSUE #217: CA Secret causes cluster to stop provisioning
/kind bug

Original author: @rbankston

https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/issues/557

Provisioned an external etcd cluster for use with a CAPV cluster and loaded the certificate authority key and certificate used for etcd in the namespace with the -ca secret with the same label. The controller fails to move on with the provisioning using the secret and instead outputs:

```
I0906 22:31:45.913692       1 kubeadmconfig_controller.go:557] reconciler "level"=0 "msg"="Creating secret for certificate"  "name"="external-ca"
E0906 22:31:45.918152       1 kubeadmconfig_controller.go:496] reconciler "msg"="unable to create cluster certificates" "error"="secrets \"external-ca\" already exists"
E0906 22:31:45.918244       1 kubeadmconfig_controller.go:218] reconciler "msg"="unable to lookup or create cluster certificates" "error"="secrets \"external-ca\" already exists" "kubeadmconfig"={"Namespace":"default","Name":"external-controlplane-0"} "machine-name"="external-controlplane-0"
```

What did you expect to happen:
Expected the secret to be used to provision the cluster

Environment:

Cluster-api-provider-vsphere version: 0.5.0
Kubernetes version: (use kubectl version): 1.15.1
OS (e.g. from /etc/os-release): Ubuntu 18.04

chuckha:
This PR extracts certificate management from the Reconciler keeping the scope of the reconciler more focused.

It fixes #217.
Related to #59 

This does not support external etcd yet because of the way kubeadm works. There will be a following PR to add support for external etcd.

This is a BREAKING API change.

I would like to see this project move to a non-library layout as this code was not intended to used as a library and we are not versioning it as such.


vincepri:
How does all of this relate to https://github.com/kubernetes-sigs/cluster-api/tree/master/util/certs?

chuckha:
@vincepri I could use those and delete (a lot) more code!

chuckha:
/hold

need to run some local tests, will update with results

ncdc:
I wouldn't call the package `internal` (I'm trying to get us to move away from generic names like internal and util). You could do `internal/certs`?

chuckha:
@ncdc the logic behind the name was that it wasn't complex enough for its own package and `certs.Certificates` stutters a bit.

What about hmm, `cluster`?

ncdc:
Not that this is a valid justification, but there is at least `context.Context` as precedent.

I'm fine with `internal/cluster` for the time being, though!

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+	// Look up each certificate as a secret and populate the certificate/key
+	for _, certificate := range c {
+		secret := &corev1.Secret{}
+		key := client.ObjectKey{
+			Name:      certificate.SecretName(cluster.Name),
+			Namespace: cluster.Namespace,
+		}
+		if err := ctrlclient.Get(ctx, key, secret); err != nil {
+			if apierrors.IsNotFound(err) {
+				kp, err := generateCACert()
```
ncdc: Do you want this function (GetCertificates) to generate if the secret isn't there? We have GetOrCreateCertificates that handles that case?

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+	spkiHash := sha256.Sum256(certificate.RawSubjectPublicKeyInfo)
+	return "sha256:" + strings.ToLower(hex.EncodeToString(spkiHash[:]))
+}
+
+// AsSecret will convert a single certificate into a Kubernetes secret.
+func (c *Certificate) AsSecret(cluster *clusterv1.Cluster, config *bootstrapv1.KubeadmConfig) *corev1.Secret {
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: cluster.Namespace,
+			Name:      fmt.Sprintf("%s-%s", cluster.Name, c.Name),
```
ncdc:
```suggestion
			Name:      c.SecretName(cluster.Name),
```

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+
+// Certificate represents a single certificate CA.
+type Certificate struct {
+	Generated bool
+	Name      string
+	*certs.KeyPair
+}
+
+// SecretName is the expected name of the Kubernetes secret the certificate is saved in.
+func (c *Certificate) SecretName(clustername string) string {
```
ncdc:
```suggestion
func (c *Certificate) SecretName(clusterName string) string {
```

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+	return fmt.Sprintf("%s-%s", clustername, c.Name)
```
ncdc:
```suggestion
	return fmt.Sprintf("%s-%s", clusterName, c.Name)
```

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+	return out, nil
+}
+
+// hashCert will calculate the sha256 of the incoming certificate.
+func hashCert(certificate *x509.Certificate) string {
+	spkiHash := sha256.Sum256(certificate.RawSubjectPublicKeyInfo)
+	return "sha256:" + strings.ToLower(hex.EncodeToString(spkiHash[:]))
+}
+
+// AsSecret will convert a single certificate into a Kubernetes secret.
```
ncdc:
```suggestion
// AsSecret converts a single certificate into a Kubernetes secret.
```

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+	Name      string
+	*certs.KeyPair
+}
+
+// SecretName is the expected name of the Kubernetes secret the certificate is saved in.
+func (c *Certificate) SecretName(clustername string) string {
+	return fmt.Sprintf("%s-%s", clustername, c.Name)
+}
+
+// Hashes hash all the certificates stored in a CA certificate.
```
ncdc:
```suggestion
// Hashes hashes all the certificates stored in a CA certificate.
```

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+		return nil, errors.Wrap(err, "unable to parse Cluster CA Certificate")
+	}
+	out := make([]string, 0)
+	for _, c := range certificates {
+		out = append(out, hashCert(c))
+	}
+	return out, nil
+}
+
+// hashCert will calculate the sha256 of the incoming certificate.
```
ncdc:
```suggestion
// hashCert calculates the sha256 of certificate.
```

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+}
+
+// Certificate represents a single certificate CA.
+type Certificate struct {
+	Generated bool
+	Name      string
+	*certs.KeyPair
+}
+
+// SecretName is the expected name of the Kubernetes secret the certificate is saved in.
```
ncdc:
```suggestion
// SecretName returns the name of the Kubernetes secret for c.
```

internal/cluster/certificates.go
chuckha: oof yes, this was my first strategy but I broke it out as you saw. 

I must have not removed this bit, thanks for the catch. I'll update and retest.

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+func (c *Certificate) SecretName(clustername string) string {
+	return fmt.Sprintf("%s-%s", clustername, c.Name)
+}
+
+// Hashes hash all the certificates stored in a CA certificate.
+func (c *Certificate) Hashes() ([]string, error) {
+	certificates, err := cert.ParseCertsPEM(c.Cert)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to parse Cluster CA Certificate")
```
ncdc: It looks like this is only used to hash the cluster ca, but we probably should have a less specific error message here, and we can wrapf where it's called to indicate that it's for the cluster ca.

internal/cluster/certificates.go
chuckha: pushing this change then adding tests to the certificates package

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	return secret
+}
+
+// AsFiles converts a slice of certificates into bootstrap files.
+func (c Certificates) AsFiles() []bootstrapv1.File {
+	clusterCA := c.GetCertificateByName(ClusterCAName)
+	etcdCA := c.GetCertificateByName(EtcdCAName)
+	frontProxyCA := c.GetCertificateByName(FrontProxyCAName)
+	serviceAccountCA := c.GetCertificateByName(ServiceAccountName)
```
ncdc: This isn't a CA

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+			Content:     string(serviceAccountCA.Key),
+		},
+	}
+}
+
+// Validate checks that all KeyPairs are valid
+func (c Certificates) Validate() error {
+	for _, certificate := range c {
+		if !certificate.KeyPair.IsValid() {
+			return errors.Errorf("CA cert material is missing cert/key for %s", certificate.Name)
```
ncdc: Might not strictly be a CA (if it's the service account cert)

controllers/kubeadmconfig_controller.go
```diff
@@ -259,24 +258,21 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
-			if config.Spec.JoinConfiguration.Discovery.BootstrapToken == nil {
-				config.Spec.JoinConfiguration.Discovery.BootstrapToken = &kubeadmv1beta1.BootstrapTokenDiscovery{}
-			}
-			config.Spec.JoinConfiguration.Discovery.BootstrapToken.CACertHashes = hashes
+	hashes, err := certificates.GetCertificateByName(internalcluster.ClusterCAName).Hashes()
+	if err != nil {
+		log.Error(err, "Unable to generate Cluster CA certificate hashes")
+		return ctrl.Result{}, err
+	}
+	if hashes != nil {
```
ncdc: If hashes **is** nil, is that ok or do we need to error?

controllers/kubeadmconfig_controller.go
```diff
@@ -259,24 +258,21 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
 	}
 
-	// Get certificates to improve security of discovery
-	certificates, err := r.getClusterCertificates(cluster)
-	if err != nil {
+	certificates := internalcluster.NewCertificates()
+	if err := certificates.GetCertificates(ctx, r.Client, cluster); err != nil {
 		log.Error(err, "unable to lookup cluster certificates")
 		return ctrl.Result{}, err
 	}
```
ncdc: Make sure we validate that all the certs are present and valid, otherwise it's an error

controllers/kubeadmconfig_controller_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/klog/klogr"
 	bootstrapv1 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
-	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/certs"
+	cluster2 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/internal/cluster"
```
ncdc: yay auto import naming? 😄

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+
+	// FrontProxyCAName is the secret name suffix for Front Proxy CA
+	FrontProxyCAName = "proxy"
+)
+
+// Certificates are the certificates necessary to bootstrap a cluster.
+type Certificates []*Certificate
+
+// NewCertificates return an initialized but empty set of CA certificates needed to bootstrap a cluster.
+func NewCertificates() Certificates {
```
ncdc: I'm wondering if this function is necessary - could we combine this with GetCertificates?

internal/cluster/certificates.go
```diff
@@ -0,0 +1,371 @@
...
+		&Certificate{Name: ClusterCAName},
+		&Certificate{Name: EtcdCAName},
+		&Certificate{Name: ServiceAccountName},
+		&Certificate{Name: FrontProxyCAName},
+	}
+}
+
+// GetCertificateByName returns a certificate by the given name.
+// This could be removed if we use a map instead of a slice to hold certificates, however other code becomes more complex.
+func (c Certificates) GetCertificateByName(name string) *Certificate {
```
ncdc: General question about naming (and ok not to make any changes): the type is Certificates, so we could have GetByName, GenerateCertificates->Generate, SaveGeneratedCertificates-> SaveGenerated, GetOrCreateCertificates-> GetOrCreate?

controllers/kubeadmconfig_controller.go
chuckha: resolved over zoom

chuckha: i should mention the resolution:

All certs must exist at this point. A check was added to make sure that is the case. If some certs are missing it is an error.

internal/cluster/certificates.go
chuckha: I like the shortened versions better :) will update

controllers/kubeadmconfig_controller.go
chuckha: This has been added as EnsureAllExist

controllers/kubeadmconfig_controller_test.go
chuckha: lol such yay. Renamed to `internalcluster`

internal/cluster/certificates.go
chuckha: Removing this function entirely in favor of `EnsureAllExist`

internal/cluster/certificates.go
```diff
@@ -0,0 +1,411 @@
...
+			}
+			return errors.WithStack(err)
+		}
+		// If a user has a badly formatted secret it will prevent the cluster from working.
+		kp, err := secretToKeyPair(secret)
+		if err != nil {
+			return err
+		}
+		if kp == nil {
+			return errors.New("SECRET IS NIL")
```
ncdc: 😄

internal/cluster/certificates.go
```diff
@@ -0,0 +1,411 @@
...
+	}
+	return secret
+}
+
+// AsFiles converts a slice of certificates into bootstrap files.
+func (c Certificates) AsFiles() []bootstrapv1.File {
+	clusterCA := c.GetByName(ClusterCAName)
+	etcdCA := c.GetByName(EtcdCAName)
+	frontProxyCA := c.GetByName(FrontProxyCAName)
+	serviceAccountCA := c.GetByName(ServiceAccountName)
```
ncdc:
```suggestion
	serviceAccountKey := c.GetByName(ServiceAccountName)
```

internal/cluster/certificates.go
chuckha: ... SHhhhhhh you didn't see annyyyything

internal/cluster/certificates.go
```diff
@@ -0,0 +1,408 @@
...
+// Certificate represents a single certificate CA.
+type Certificate struct {
+	Generated bool
+	Name      string
+	KeyPair   *certs.KeyPair
+}
+
+// SecretName is the expected name of the Kubernetes secret the certificate is saved in.
+func (c *Certificate) SecretName(clusterName string) string {
+	return fmt.Sprintf("%s-%s", clusterName, c.Name)
```
fabriziopandini: What about calling [this](https://github.com/kubernetes-sigs/cluster-api/blob/9bb49677bc1d8cf97cfea09d1cb0dc3b6d8fd230/util/secret/secret.go#L45) instead of duplicating this func?

internal/cluster/certificates.go
```diff
@@ -0,0 +1,408 @@
...
+		secret := certificate.AsSecret(cluster, config)
+		if err := ctrlclient.Create(ctx, secret); err != nil {
+			return errors.WithStack(err)
+		}
+	}
+	return nil
+}
+
+// GetOrCreate is a convenience function that wraps cluster bootstrap certificate behavior.
+func (c Certificates) GetOrCreate(ctx context.Context, ctrlclient client.Client, cluster *clusterv1.Cluster, config *bootstrapv1.KubeadmConfig) error {
```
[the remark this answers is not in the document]
chuckha: that's fine with me

internal/cluster/certificates.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+// hashCert calculates the sha256 of certificate.
+func hashCert(certificate *x509.Certificate) string {
+	spkiHash := sha256.Sum256(certificate.RawSubjectPublicKeyInfo)
+	return "sha256:" + strings.ToLower(hex.EncodeToString(spkiHash[:]))
+}
+
+// AsSecret converts a single certificate into a Kubernetes secret.
+func (c *Certificate) AsSecret(cluster *clusterv1.Cluster, config *bootstrapv1.KubeadmConfig) *corev1.Secret {
```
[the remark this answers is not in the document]
chuckha: We certainly could. Would it be ok to do this in a follow up issue so we can discuss behavior expectations?

It might be worth mentioning this package is internal so no one outside of CABPK will be able to use it, but I'm sure it would still be more friendly to return an error or produce an empty secret.

internal/cluster/certificates.go
chuckha: it's reasonable. I think there's some way to instantiate a `Certificates` better than this, but I'm not sure exactly what that looks like yet. Do you think we could keep it like this for now until we come up with something better? Instantiation of the four different KeyPairs doesn't quite feel right to me inside the Lookup method, but it could be argued either way. If you feel strongly I can spend some time figuring something else out though

internal/cluster/certificates.go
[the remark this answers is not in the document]
fabriziopandini: Sure things!

fabriziopandini:
/approve
Pending for @ncdc answer to [comment](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/222#discussion_r324981270) before lgtm

internal/cluster/certificates.go
ncdc: Fine to keep this for now

ncdc:
@fabriziopandini over to you for lgtm (the f2f is keeping me busy) - thanks!

PR #226, opened 2019-09-13
:bug: Fix error-return for machine-cluster association

### LINKED ISSUE #212: should return without an error if no cluster is set on the machine
/kind bug

https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/cc2b1e48885f2f7bda4b8a3048b1c33367322a0f/controllers/kubeadmconfig_controller.go#L128-L132

As of today the reconciler returns an error if no cluster can be found for a machine. Instead, we should either requeue or simply return and try again later.

If there is no label `utils.GetClusterFromMetadata` returns a specific error. In that case, we should return no error. Perhaps there should be a log line indicating there is no associated cluster yet? The reason we should not error is that If the machine gets updated with a cluster the associated KubeadmConfig will get reconciled anyway and can proceed past this point.

If there *is* a label but the cluster does not exist in etcd yet we should also return. This assumes that the cluster *will* be created. When the cluster is created the associated KubeadmConfigs will all reconcile and the reconciler can proceed past this point.

The only time this should return an error is when there is a real error getting data from the API server.

Thoughts on this one @detiber ?

kashifest:
What this PR does / why we need it:
Removes error and waits again if

a) machine does not have any associated cluster
b) Machine has associate cluster but it is still non-existent

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #212

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -127,6 +127,16 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 	// Lookup the cluster the machine is associated with
 	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, machine.ObjectMeta)
 	if err != nil {
+		if errors.Cause(err) == util.ErrNoCluster {
+			log.Info("Machine does not belong to a cluster yet, waiting until its part of a cluster")
+			return ctrl.Result{}, nil
+		}
+
+		if apierrors.IsNotFound(err) {
+			log.Info("Cluster does not exist yet , waiting until it is created")
+			return ctrl.Result{}, nil
+		}
+
 		log.Error(err, "could not get cluster by machine metadata")
 		return ctrl.Result{}, err
 	}
--- a/controllers/kubeadmconfig_controller_test.go
+++ b/controllers/kubeadmconfig_controller_test.go
@@ -152,9 +152,8 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasBootstrapData(
 	}
 }
 
-// This returns an error since nothing can proceed without an associated cluster.
-// TODO: This should probably not return an error
-func TestKubeadmConfigReconciler_Reconcile_ReturnErrorIfMachineDoesNotHaveAssociatedCluster(t *testing.T) {
+// This does not expect an error, hoping the machine gets updated with a cluster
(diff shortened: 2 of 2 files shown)


controllers/kubeadmconfig_controller.go
```diff
@@ -127,8 +127,16 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
+		if errors.Cause(err) == util.ErrNoCluster {
+			log.Info("Machine does not belong to a cluster yet, waiting until its part of a cluster")
+			return ctrl.Result{}, nil
+		} else if apierrors.IsNotFound(err) {
+			log.Info("Cluster does not exist yet , waiting until it is created")
+			return ctrl.Result{}, nil
+		} else {
+			log.Error(err, "could not get cluster by machine metadata")
+			return ctrl.Result{}, err
+		}
```
detiber: I'm not seeing a clean way to switch this into a switch statement to appease the linter, but since each branch has a return, you can remove the else clauses:

```go
		if errors.Cause(err) == util.ErrNoCluster {
			log.Info("Machine does not belong to a cluster yet, waiting until its part of a cluster")
			return ctrl.Result{}, nil
		}

		if apierrors.IsNotFound(err) {
			log.Info("Cluster does not exist yet , waiting until it is created")
			return ctrl.Result{}, nil
		}

		log.Error(err, "could not get cluster by machine metadata")
		return ctrl.Result{}, err		
```

kashifest: @detiber Thanks. I will do the change

controllers/kubeadmconfig_controller.go
```diff
@@ -127,6 +127,16 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 	// Lookup the cluster the machine is associated with
 	cluster, err := util.GetClusterFromMetadata(ctx, r.Client, machine.ObjectMeta)
 	if err != nil {
```
chuckha: What do y'all think about a switch statement instead?

```
switch {
case errors.Cause(err) == util.ErrNoCluster:
case apierrors.IsNotFound(err):
case err != nil:
}
```

?

chuckha:
/approve

if you end up using the switch that's cool, but i'm happy with this as is also!

PR #230, opened 2019-09-16
:sparkles: Add lint to make test command

kashifest:
Adds lint test to make test command


chuckha:
an improvement for developers for sure

/approve

vincepri:
Pinging @ncdc, which IIRC has some thoughts around adding more pre-reqs to make targets

chuckha:
Yes, I've got lots of thoughts too, but we should put them down in an issue first. I don't think this PR is the place to discuss broader improvements. I don't want to block this PR on us figuring out how to clean up the makefile vs hack/verify scripts problem.

PR #232, opened 2019-09-17
🐛 Fix init lock

accepting:

Fix bug for the init lock. 

Consider the following case:
Two control plane machines(note them as C1, C2) are creating concurrently. Assume C1 first acquires the init lock. Since C2 failed to get lock, it exits the reconcile loop and releases the init lock. 
**Before C1 finished to initialize itself**, we create another control plane machine(note it as C3). Obviously, C3 will acquire the init lock successfully and also generate controlplane-init config which is what we do not want to see.

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -162,13 +162,6 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 		}
 	}()
 
-	holdLock := false
-	defer func() {
-		if !holdLock {
-			r.KubeadmInitLock.Unlock(ctx, cluster)
-		}
-	}()
-
 	if !cluster.Status.ControlPlaneInitialized {
 		// if it's NOT a control plane machine, requeue
 		if !util.IsControlPlaneMachine(machine) {
@@ -189,6 +182,12 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 			log.Info("A control plane is already being initialized, requeing until control plane is ready")
 			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 		}
+
+		defer func() {
+			if rerr != nil {
+				r.KubeadmInitLock.Unlock(ctx, cluster)
+			}
+		}()
 
 		log.Info("Creating BootstrapData for the init control plane")
 
@@ -253,8 +252,6 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 
 		config.Status.BootstrapData = cloudInitData
 		config.Status.Ready = true
-
-		holdLock = true
 
 		return ctrl.Result{}, nil
 	}


chuckha:
The second control plane can't release the lock that the first control plane is holding. If it *does* release the lock that is a bug.

vincepri:
Reading this code https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/master/internal/locking/control_plane_init_mutex.go#L100-L126 it seems like we release the lock without checking if the Machine that has the lock is the same as the current one?

vincepri:
Which is something we do here https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/master/internal/locking/control_plane_init_mutex.go#L65-L73

controllers/kubeadmconfig_controller.go
```diff
@@ -162,14 +162,12 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
-	defer func() {
-		if !holdLock {
-			r.KubeadmInitLock.Unlock(ctx, cluster)
-		}
-	}()
-
 	if !cluster.Status.ControlPlaneInitialized {
+		defer func() {
+			if rerr != nil {
+				r.KubeadmInitLock.Unlock(ctx, cluster)
```
chuckha: There is another Unlock call down beneath this code block that we can delete: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/master/controllers/kubeadmconfig_controller.go#L265

chuckha:
@vincepri !!! yeah that would be a bug 😞

accepting:
@vincepri @chuckha I don't think it's a good idea that only the owner machine can release the lock.  As far as I know, the lock should be hold from the first acquiring to the time when `cluster.Status.ControlPlaneInitialized` turns to be true. Therefore, the lock owner maybe have no chance to release it.

controllers/kubeadmconfig_controller.go
accepting: Without this defer block, the first owner machine of the lock will always hold the lock even if it occurs errors during generating cloud init for bootstrap control plane.

chuckha: Agreed. The defer block must stay. I was suggesting deleting the Unlock call on line 265. Do not remove the defer block! It is necessary!

chuckha: Oh, but rereading the code I'm wrong. The unlock on line 265 has to stay otherwise the lock is held forever.

accepting: Yes, both of them are necessary.

chuckha:
In the easy case when there is simply one control plane machine:

1. The control plane machine acquires the lock
2. the control plane machine does not error out and sets bootstrap data.  the lock is held
3. some time passes and the control plane becomes initialized at which time the kubeadm will be reconciled again (due to the cluster watch)
4. the control plane machine returns early because it has bootstrap data set.

Now a new bootstrap config is created for a worker machine.
1. the control plane is initialized so the the lock is released. Not that it matters at this point.

Let's say there are two control planes and two managers. This situation should never happen since the deployment uses leader election to ensure only a single manager is running. But let's say this does happen.

The case you originally outlined indeed holds true. C1 acquires the lock. C2 attempts to acquire the lock and cannot and errors out and thus removes the lock. C3 is created and acquires the lock before C1 finishes initializing. Now you have two clusters. Very bad.

What if we never released the lock in the successful case?

I think one clean way to do this is to acquire the lock and never unlock it. The only time we want to unlock the lock is if we hit an error condition after the lock is created but before the bootstrap data is generated.

This simplifies the code a bit. This change is an improvement but we should still delete the Unlock code on line 265.

Interested to hear your thoughts @accepting

accepting:
@chuckha In my opinion, the exact place to release the lock is where `cluster.Status.ControlPlaneInitialized` is set to true in the cluster controller.  If we can not make the cluster controller do the work, **the Unlock code on line 265** may be the most right place in CABPK. 

The only bad effect I can see is that **the Unlock code on line 265** will be involved every time a new bootstrap config is created afterwards. That is really unnecessary.

If we never released the lock in the successful case, it can avoid the bad effect above. But It may lead some bad effects. For example,
1. Create a cluster with only one control plane machine, note it as C1.
2. C1 acquires the init lock successfully and hold it forever. That means the configmap `{cluser-namespace}/{cluster-name}-lock` always exist. And the configmap contains C1's machine name.
3. After some days, we realize some problems in this cluster and want to rebuild this cluster.  We delete the cluster and all machines in this cluster.
4. We recreate the cluster **with the same cluster name** and **with different control planes**.
5. Create control plane NC1 and its bootstrap config. NC1 tries to acquire the init lock. At this time, since the configmap `{cluser-namespace}/{cluster-name}-lock` exists, and the information in the configmap is C1's machine name which not equals to NC1's machine name, NC1 fails to acquire the init lock. Hence, NC1 can not generate bootstrap data forever.

chuckha:
@accepting the config map has an owner reference so it will be deleted when we delete the owning machine

controllers/kubeadmconfig_controller.go
```diff
@@ -162,14 +162,12 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
-	holdLock := false
-	defer func() {
-		if !holdLock {
-			r.KubeadmInitLock.Unlock(ctx, cluster)
-		}
-	}()
-
 	if !cluster.Status.ControlPlaneInitialized {
+		defer func() {
+			if rerr != nil {
```
chuckha: I'm interested in this change, why do we only unlock when rerr has an error?

chuckha:
another way this would work is:

If the bootstrap config already has bootstrap data
and the bootstrap config is holding the lock we can then delete the lock at that point

accepting:
@chuckha You are right, I missed that. But the owner reference is set as the cluster now. We may need to change it to the related machine.

controllers/kubeadmconfig_controller.go
accepting: Since any error case in this block means the bootstrap data is not generated successfully, we restore the lock. Do you have any better suggestion?

accepting: @chuckha I have moved this code behide the Lock code.

PR #234, opened 2019-09-17
✨ Initialize JoinConfiguration if missing

### LINKED ISSUE #213: Should not error if worker node does not have JoinConfiguration
/kind bug

https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/cc2b1e48885f2f7bda4b8a3048b1c33367322a0f/controllers/kubeadmconfig_controller.go#L258-L260

After the control plane is initialized and a worker gets created but that worker does not have a JoinConfiguration the reconciler returns an error.

I expected a nil join configuration would be absolutely fine in this case. The reconciler should imply set it to an initialized JoinConfiguration. If we detect that a Cluster or Init configuration is set then we should log a line saying "the worker will ignore a set Cluster/Init configuration" and proceed with the join. Right now we are ignoring Cluster/Init without notifying the user which may lead to confusion.

fabriziopandini:
- If join configuration is missing, we are going to generate a default one.

**Which issue(s) this PR fixes**:
Fixes #213


accepting:
@fabriziopandini  In our use cases, we use a template to generate bootstrap configs for all control plane machines. Hence, these configs all contains initconfiguration, clusterconfiguration and joinconfiguration. If adding the requirements in your PR, it will make the work of setting kubeadmconfig objects more manual.

fabriziopandini:
@accepting [here](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/213#issuecomment-528869153) a thread explaining why it was decided to make this change

chuckha:
I'm really torn on this PR

I'm leaning towards not being restrictive to allow more use cases, but I'm concerned about how people are using this. I don't think I have enough information about what is more common, using duplicated configs across all machines or being specific about which machine will be the initial control plane and which will be a secondary control plane. I'll talk to some people this week and see if I can get some consensus on this.

Any additional thoughts are welcome.

Please see the associated issue for more details. The TL;DR is: Should we allow users to supply all configs to every machine and let the controller figure it out? or should we be prescriptive to prevent accidental user errors and nice error messages.

chuckha:
For this PR can we have just the initializing a join configuration if one doesn't exist and leave the other checks out for now? That way our behavior doesn't change and we can close #213.

SataQiu:
/retitle ✨ Initialize config objects if missing

controllers/kubeadmconfig_controller.go
```diff
@@ -256,13 +256,21 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 	}
 
 	// Every other case it's a join scenario
-	// Nb. in this case ClusterConfiguration and JoinConfiguration should not be defined by users, but in case of misconfigurations, CABPK simply ignore them
+
+	//In case of join, ClusterConfiguration and InitConfiguration should not be defined by users
+	if config.Spec.ClusterConfiguration != nil || config.Spec.InitConfiguration != nil {
+		return ctrl.Result{}, errors.New("Control plane already exists for the cluster, KubeadmConfig objects with ClusterConfiguration or InitConfiguration can not be processed at this stage")
```
detiber: It might be good to also emit an event here as well.

ncdc: If we return an error here, it means the KubeadmConfig will be reconciled again using backoff. We probably don't want to return an error.

chuckha: I asked @fabriziopandini to not make these changes as they are controversial and reduce this PR to fix the original issue of #213

fabriziopandini:
@chuckha 
Reduced the scope to 
- If join configuration is missing, we are going to generate a default one.

controllers/kubeadmconfig_controller.go
```diff
@@ -256,13 +258,18 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
 	// Unlock any locks that might have been set during init process
 	r.KubeadmInitLock.Unlock(ctx, cluster)
 
+	// if the JoinConfiguration is missing, create a default one
 	if config.Spec.JoinConfiguration == nil {
-		return ctrl.Result{}, errors.New("Control plane already exists for the cluster, only KubeadmConfig objects with JoinConfiguration are allowed")
+		log.Info("Creating default JoinConfiguration")
+		config.Spec.JoinConfiguration = &kubeadmv1beta1.JoinConfiguration{}
+		if util.IsControlPlaneMachine(machine) {
+			config.Spec.JoinConfiguration.ControlPlane = &kubeadmv1beta1.JoinControlPlane{}
```
detiber: Shouldn't this be dependent on if it is a ControlPlane instance or a Worker Node instance?

chuckha: i think you missed line 270's if check

chuckha:
/hold cancel

looks good to me. assigning to @detiber in case I've misunderstood his comment

/assign @detiber

controllers/kubeadmconfig_controller.go
detiber: It appears I did :(

PR #236, opened 2019-09-18
✨ using golangci-lint instead of verify-xxx.sh

### LINKED ISSUE #223: Extract golint into a tool in hack/tools
/kind feature

The golint tool is used to lint our codebase. It's being installed in a tempdir but it should follow the existing pattern of tools.

The hack/verify-golint.sh file should not be doing any of this:

https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/23205b40552da0c99f822458aa02a97df583d5e4/hack/verify-golint.sh#L39-L50

Perhaps it's entirely replaced with golangci-lint?

https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/tree/master/hack/tools

That is for the ticket owner to decide. 
/priority important-soon
/milestone v0.1.x

SataQiu:
using golangci-lint instead of verify-xxx.sh 

Fixes #223

/cc @chuckha


chuckha:
/approve

very nice!

hack/verify-all.sh
```diff
@@ -73,13 +73,6 @@ if [[ "${VERIFY_GOFMT:-true}" == "true" ]]; then
...
 fi
 
-if [[ "${VERIFY_GOLINT:-true}" == "true" ]]; then
-  echo "[*] Verifying golint..."
-  out=$(hack/verify-golint.sh 2>&1)
-  failure $? "verify-golint.sh" "${out}"
-  cd "${REPO_PATH}"
-fi
-
 if [[ "${VERIFY_GOVET:-true}" == "true" ]]; then
```
detiber: The GOLANGCI_LINT section below should already run govet, so this can be removed (along with the `hack/verify-govet.sh` script.

hack/verify-golint.sh
```diff
@@ -18,34 +18,5 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-# shellcheck source=/dev/null
```
detiber: You should be able to remove this whole file, since the invocation of `verify-golangci-lint` should already invoke the linters we are requesting here.

hack/verify-all.sh
```diff
@@ -73,13 +73,6 @@ if [[ "${VERIFY_GOFMT:-true}" == "true" ]]; then
   cd "${REPO_PATH}"
```
detiber: The GOLANGCI_LINT section below already runs gofmt, so the VERIFY_GOFMT section can be removed (along with the hack/verify-gofmt.sh script.

detiber: The GOLANGCI_LINT section below already runs misspell, so the VERIFY_SPELLING section can be removed (along with the hack/verify-spelling.sh script.

detiber: The GOLANGCI_LINT section below already runs whitespace, so the VERIFY_WHITESPACE section can be removed (along with the hack/verify-whitespace.sh script.

Makefile
```diff
@@ -101,6 +101,14 @@ lint: $(GOLANGCI_LINT) ## Lint quickly using `golangci-lint --fast=true`
 lint-full: $(GOLANGCI_LINT) ## Lint thoroughly using `golangci-lint --fase=false`
 	$(GOLANGCI_LINT) run -v --fast=false
 
+.PHONY: lint-focus
+lint-focus: $(GOLANGCI_LINT) ## Lint using specified linter
```
detiber: Since `hack/verify-all.sh` is already calling one of the other make targets, this target shouldn't be needed.

SataQiu: Thanks so much @detiber 
Actually, I've been thinking about doing this for a long time.
After doing this, I feel the code is much cleaner :)

PR #240, opened 2019-09-20
🏃Added Test :Return early if the owning machine does not have an associated cluster

hpandeycodeit:
This PR adds a test case  - Return early if the owning machine does not have an associated cluster.

**Which issue(s) this PR fixes**  :
Ref #214

### DIFF
--- a/controllers/kubeadmconfig_controller_test.go
+++ b/controllers/kubeadmconfig_controller_test.go
@@ -192,6 +192,34 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasBootstrapData(
 	}
 }
 
+// Return early If the owning machine does not have an associated cluster
+func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasNoCluster(t *testing.T) {
+	machine := newMachine(nil, "machine") // Machine without a cluster
+	config := newKubeadmConfig(machine, "cfg")
+
+	objects := []runtime.Object{
+		machine,
+		config,
+	}
+	myclient := fake.NewFakeClientWithScheme(setupScheme(), objects...)
+
+	k := &KubeadmConfigReconciler{
+		Log:    log.Log,
+		Client: myclient,
+	}
+
+	request := ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: "default",
+			Name:      "cfg",
+		},
+	}
+	_, err := k.Reconcile(request)
+	if err != nil {
+		t.Fatalf("Not Expecting error, got an error: %+v", err)
+	}
+}
+
 // This does not expect an error, hoping the machine gets updated with a cluster
 func TestKubeadmConfigReconciler_Reconcile_ReturnNilIfMachineDoesNotHaveAssociatedCluster(t *testing.T) {
 	machine := newMachine(nil, "machine") // intentionally omitting cluster


controllers/kubeadmconfig_controller_test.go
```diff
@@ -192,6 +192,34 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasBootstrapData(
 	}
 }
 
+// Return early If the owning machine does not have a associated cluster
+func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasNoCluster(t *testing.T) {
```
detiber: @chuckha @ncdc based on the discussions from the f2f planning, is this the behavior that we actually want? Or do we want to allow a Machine without a Cluster as long as the label isn't set and have some fallback behavior to just not provide the defaults we set from the Cluster resource when that is the case?

controllers/kubeadmconfig_controller_test.go
```diff
@@ -192,6 +192,34 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasBootstrapData(
 	}
 }
 
+// Return early If the owning machine does not have a associated cluster
```
SataQiu:
```suggestion
// Return early If the owning machine does not have an associated cluster.
```

controllers/kubeadmconfig_controller_test.go
chuckha: This is for the use-case that only uses the Machine API?

detiber: @chuckha indeed

chuckha: My understanding was that a Cluster object was still required even if it's an empty object. It's good to figure this out now though. Is that also your understanding?

detiber: That is the direction that @ncdc, @vincepri, and myself have been advocating for, but I believe that the Cluster has been specified to still be optional at this point in time.

chuckha: Cluster is required as of today. This is not a change in behavior but a formalization of the current behavior. If we can open an issue to get the Machine API use-case here then we can work on iterating from this place:

[CABPK requires a Cluster](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/c94ead1dbfc947ed7625d2eb3c8b4183ac7efd62/controllers/kubeadmconfig_controller.go#L128)
[Cluster API requires a Cluster](https://github.com/kubernetes-sigs/cluster-api/blob/2c4cbfa482d3bc7857333aeb1a0eb0599e2320bd/controllers/machine_controller.go#L113)
[CAPA requires a Cluster](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/256276cd93bb38d4a1d6dae98d376908689c9d5d/controllers/awsmachine_controller.go#L91)

controllers/kubeadmconfig_controller_test.go
```diff
@@ -192,6 +192,34 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasBootstrapData(
...
+	}
+
+	request := ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: "default",
+			Name:      "cfg",
+		},
+	}
+	_, err := k.Reconcile(request)
+	if err == nil {
```
chuckha: I believe this recently changed, If the cluster does not exist then we expect error to be nil as it will simply be reconciled again. If you flip the `==` to be `!=` and fix up the t.Fatal call then this looks good to me

hpandeycodeit: done

controllers/kubeadmconfig_controller_test.go
hpandeycodeit: done.

chuckha:
can you please squash your commits?

controllers/kubeadmconfig_controller_test.go
```diff
@@ -192,6 +192,34 @@ func TestKubeadmConfigReconciler_Reconcile_ReturnEarlyIfMachineHasBootstrapData(
...
+
+	request := ctrl.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: "default",
+			Name:      "cfg",
+		},
+	}
+	_, err := k.Reconcile(request)
+	if err != nil {
+		t.Fatal("Not Expecting error, got an error")
```
chuckha: It would be good to also print out the error, something like `t.Fatalf("Not Expecting error, got an error: %+v", err)`

hpandeycodeit: thanks for pointing it out! Added the error in the t.fatalf

PR #241, opened 2019-09-22
set init lock's owner as the holding machine

accepting:

Since the init lock holds information about the owner machine, it is better to set the configmap's OwnerReference as the holding machine. Then the configmap can be automatically deleted when we delete the owner machine.

### DIFF
--- a/internal/locking/control_plane_init_mutex.go
+++ b/internal/locking/control_plane_init_mutex.go
@@ -76,7 +76,7 @@ func (c *ControlPlaneInitMutex) Lock(ctx context.Context, cluster *clusterv1.Clu
 	}
 
 	// Adds owner reference, namespace and name
-	sema.setMetadata(cluster)
+	sema.setMetadata(cluster, machine)
 	// Adds the additional information
 	if err := sema.setInformation(&information{MachineName: machine.Name}); err != nil {
 		log.Error(err, "Failed to acquire lock while setting semaphore information")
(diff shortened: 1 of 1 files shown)


detiber:
/assign @chuckha 

I can't think of any reason that this wouldn't work with the current way we are using the lock today, but I'm not necessarily sure that we couldn't ensure that we don't introduce ways that this couldn't possibly come back to bite us future

chuckha:
@accepting In future if you could address why and not just what in your PRs that would be really helpful to reviewers. Alternatively filing an issue beforehand can also help to set context.

Some additional context for this issue:

The reason the machine owns the lock is that the lock is based on machine. The first control plane actually initializes the cluster, not the Cluster object.

If the cluster owns the lock it will not be removed until the Cluster object is deleted, but we don't have to delete the cluster object. If Machines are improperly configured and fail to initialize the lock will not get released appropriately. If we then delete the machines the lock remains even though it has no more meaning. If we create new machines in the same cluster we end up in a weird state.

chuckha:
Ah, so one thing @detiber pointed out is that Cluster.Status.ControlPlaneReady won't be reset when machines get deleted. The original use case was deleting machines but not the control plane to recreate the cluster. How is your use case handling the `Cluster.Status.ControlPlaneReady` field?

ncdc:
FYI, it's `ControlPlaneInitialized` (we wanted to be explicit that it was a one-time toggle from false to true, rather than an indication that we're keeping the status constantly up to date. Although I guess if you delete all the control plane machines, the value is incorrect at that point, unless we flip it back to false).

@chuckha aren't we still deleting the configmap as soon as we've successfully initialized the first control plane machine?

chuckha:
@ncdc yes, that's right, we delete the lock if the control plane is initialized successfully and another machine attempts to reconcile.

The config map is not deleted if `kubeadm init` fails to succeed.

chuckha:
It almost feels like Cluster should be renamed to Infrastructure and the cluster concept should become ControlPlane...but we'll hammer that idea out during v1a3 planning

chuckha:
@accepting we can't merge this as it breaks the `ControlPlaneInitialized` bool on cluster.

If ControlPlaneInitialized is true it can't go back to false for this version of Cluster API. Therefore the lock must be owned by the Cluster object.

/close

PR #242, opened 2019-09-22
Patch kubeadmconfig objects only if no error occurs during reconciliation

accepting:

There is no need to update KubeadmConfig objects if any error occurs during the reconciliation.

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -151,12 +151,11 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
 	if err != nil {
 		return ctrl.Result{}, err
 	}
-	// Always attempt to Patch the KubeadmConfig object and status after each reconciliation.
+	// Attempt to Patch the KubeadmConfig object and status after each reconciliation if no error occurs.
 	defer func() {
-		if err := patchHelper.Patch(ctx, config); err != nil {
-			log.Error(err, "failed to patch config")
-			if rerr == nil {
-				rerr = err
+		if rerr == nil {
+			if rerr = patchHelper.Patch(ctx, config); rerr != nil {
+				log.Error(rerr, "failed to patch config")
 			}
 		}
 	}()


controllers/kubeadmconfig_controller.go
```diff
@@ -151,12 +151,11 @@ func (r *KubeadmConfigReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, re
...
 		return ctrl.Result{}, err
 	}
-	// Always attempt to Patch the KubeadmConfig object and status after each reconciliation.
+	// Attempt to Patch the KubeadmConfig object and status after each reconciliation if no error occurs.
 	defer func() {
-		if err := patchHelper.Patch(ctx, config); err != nil {
-			log.Error(err, "failed to patch config")
-			if rerr == nil {
-				rerr = err
+		if rerr == nil {
```
detiber: This seems like a departure from other cluster-api related use cases, where we would potentially want to ensure that we persist some type of intermediate or partial state back to the resource. That said, the behaviors of this controller vs those other cases it is probably generally ok.

Are we attempting to set ErrorMessage and ErrorReason anywhere where this change could potentially prevent those changes from persisting?

chuckha: We do not currently use ErrorMessage or ErrorReason in this provider

chuckha:
@accepting This is right. We never set the config value except in the successful case so there is no reason to issue a patch in an error case, but I'm curious why you want this change? Was there a bug you were encountering that this fixes?

accepting:
@chuckha In our use cases, the bootstrap objects support update. When we try to update a bootstrap object, we found the object was updated even if some error occurs afterwards.

chuckha:
That's interesting. Does an update to the bootstrap object actually modify the cluster?

accepting:
Yes, with the help of kubeadm phases, we update the control planes/worker nodes' settings by updating their bootstrap objects.

chuckha:
I'm fine merging this and reverting or changing behavior back if we need to start populating the error message.

/approve

PR #251, opened 2019-09-25
🏃 Add test for patch behavior

### LINKED ISSUE #243: Add a test for patch behavior
/kind feature

There should be a test that tests the behavior of saving an object.

The expected behavior is that if there is an error in reconcile at any point no Patch will be issued.

If there is no error in Reconcile, then a Patch will be issued.

This is something of a tricky test which is why I've not marked it as good first issue. If you really like testing it might be an ok first issue.

/priority important-soon
/milestone v0.1.x

SataQiu:
Add test for patch behavior

Fixes #243


chuckha:
/approve
/lgtm

🙌  thank you!

PR #256, opened 2019-09-27
🏃Add release notes utility / update controller-tools & CRDs

vincepri:

This PR adds the `make release-notes` target

Fixes #


go.mod
```diff
@@ -4,19 +4,27 @@ go 1.12
...
+	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
 	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
 	k8s.io/cluster-bootstrap v0.0.0-20190516232516-d7d78ab2cfe7
 	k8s.io/klog v0.4.0
-	k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208 // indirect
+	k8s.io/kube-openapi v0.0.0-20190816220812-743ec37842bf // indirect
 	sigs.k8s.io/cluster-api v0.2.0
 	sigs.k8s.io/controller-runtime v0.2.0
```
detiber: Would it be worth it to rev controller-runtime and cluster-api while in here?

vincepri: I'll follow-up here, these usually come when we update CAPI

chuckha:
oh boy, the tools controller-runtime is different than the actual controller-runtime dev 🙃 

they should be kept in sync

hack/tools/go.mod
```diff
@@ -4,5 +4,6 @@ go 1.12
 
 require (
 	github.com/golangci/golangci-lint v1.18.0
-	sigs.k8s.io/controller-tools v0.2.0
+	sigs.k8s.io/cluster-api/hack/tools v0.0.0-20190926205936-6fc621a1c6af
+	sigs.k8s.io/controller-tools v0.2.1
```
chuckha: is this change necessary?

vincepri: It comes with CAPI, yes

chuckha: we're only importing tools for the release notes which does not depend on controller-tools

This change is unnecessary for this specific PR, but if you want to bump this version now you will also have to bump it in go.mod or change the CRD verification test

vincepri: `sigs.k8s.io/cluster-api/hack/tools` imports `sigs.k8s.io/controller-tools` v0.2.1, which is why it gets updated

chuckha: yeah, i saw that, but that doesn't mean we have to bump this version, the tools in this repo are still building on v0.2.0

vincepri: The controller-tools are only used in `hack/tools` as far as I can tell. Are you referring to the controller-runtime version?

chuckha: sorry, let me try again:

the crds are built (or, at least verified) using controller-tools v0.2.0
This update is unnecessary for the scope of this PR. Just because capi/tools uses controller-tools v0.2.0 does not mean we've made the update in this repo.

Can you update the title/description to include the CRD updates if the controller-tools dependency is also getting changed?

vincepri: ok reworded, and split the commits, ptal

detiber:
/lgtm
/hold
holding for @chuckha

PR #257, opened 2019-09-27
🏃Align Makefile with other CAP* repos

vincepri:

Fixes #


vincepri:
/hold

until #256 merges

Makefile
```diff
@@ -192,17 +192,39 @@ release-manifests: $(RELEASE_DIR) ## Builds the manifests to publish with a rele
...
 .PHONY: clean-release
 clean-release: ## Remove the release folder
 	rm -rf $(RELEASE_DIR)
+
+## --------------------------------------
+## Others / Utilities
+## --------------------------------------
+
+.PHONY: run
+run: generate fmt vet ## Run against the configured Kubernetes cluster in ~/.kube/config
```
chuckha: it looks like fmt and vet get removed in this change but are now referenced here

vincepri: Yeah those are part of golint, I'll remove these

chuckha:
/hold

looks like fmt and vet snuck back in but they are still deleted?

Makefile
chuckha: these were gone but they look to be back?

PR #260, opened 2019-09-27
✨ Adds support for external etcd & respects custom certificates directory from kubeadm

### LINKED ISSUE #59: Improve user experience when using an external etcd

### LINKED ISSUE #255: Does not support a custom pki directory
/kind bug

Kubeadm allows for [custom CertificatesDir](https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/master/kubeadm/v1beta1/types.go#L107), but we hardcode it to `/etc/kubernetes/pki`.

I expect CABPK to respect my kubeadm config.

this is not a super common feature from my experience in kubeadm but we should support it as people will expect this to work

/priority important-longterm
/milestone Next

chuckha:

This PR adds additional support + documentation for creating kubernetes clusters that use external etcd.

cc @rbankston

Fixes #59
Fixes #255

### DIFF
--- a/docs/external-etcd.md
+++ b/docs/external-etcd.md
@@ -1,0 +1,77 @@
+# Support for external etcd
+
+Cluster API Bootstrap Provider Kubeadm  supports using an external etcd cluster for your workload Kubernetes clusters.
+
+### ⚠️ Warnings ⚠️
+
+Before getting started you should be aware of the expectations that come with using an external etcd cluster.
+
+* Cluster API is unable to manage any aspect of the external etcd cluster.
+* Depending on how you configure your etcd nodes you may incur additional cloud costs in data transfer.
+    * As an example, cross availability zone traffic can cost money on cloud providers. You don't have to deploy etcd
+    across availability zones, but if you do please be aware of the costs.
+
+### Getting started
+
+To use this, you will need to create an etcd cluster and generate an apiserver-etcd-client key/pair.
+[`etcdadm`](https://github.com/kubernetes-sigs/etcdadm) is a good way to get started if you'd like to test this
+behavior.
+
+Once you create an etcd cluster, you will want to base64 encode the `/etc/etcd/pki/apiserver-etcd-client.crt`,
+`/etc/etcd/pki/apiserver-etcd-client.key`, and `/etc/etcd/pki/server.crt` files and put them in two secrets. The secrets
+must be formatted as follows and the cert material must be base64 encoded:
+
+```yaml
+# Kubernetes APIServer etcd client certificate
+kind: Secret
+apiVersion: v1
+metadata:
+  name: $CLUSTER_NAME-apiserver-etcd-client
+  namespace: $CLUSTER_NAMESPACE
+data:
+  tls.crt: |
+    LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSURCRENDQWV5Z0F3SUJBZ0lJZFlkclZUMzV0
+    NW93RFFZSktvWklodmNOQVFFTEJRQXdEekVOTUFzR0ExVUUKQXhNRVpYUmpaREFlRncweE9UQTVN
+    ...
+  tls.key: |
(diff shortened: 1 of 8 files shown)


internal/cluster/certificates.go
```diff
@@ -67,13 +74,65 @@ var (
...
+	}
+
+	certificates = append(certificates, etcdCert)
+	return certificates
+}
+
+// NewCertificatesForWorker return an initialized but empty set of CA certificates needed to bootstrap a cluster.
+func NewCertificatesForWorker(caCertPath string) Certificates {
+	if caCertPath == "" {
+		caCertPath = defaultCertificatesDir + "ca.crt"
```
SataQiu: Thanks @chuckha :+1: 
But I prefer to write like this 
```suggestion
		caCertPath = filepath.Join(defaultCertificatesDir, "ca.crt")
```
WDYT ?

chuckha: Yes, this is objectively better

docs/external-etcd.md
```diff
@@ -0,0 +1,67 @@
+# Support for external etcd
```
vincepri: We should embed this an advanced section in the CAPI book

chuckha: :D yes!

docs/external-etcd.md
```diff
@@ -0,0 +1,67 @@
...
+
+Cluster API Bootstrap Provider Kubeadm now supports using an external etcd cluster for your workload Kubernetes
+clusters.
+
+To use this you will need to create an etcd cluster and generate an apiserver-etcd-client key/pair.
+[`etcdadm`](https://github.com/kubernetes-sigs/etcdadm) is a good way to get started if you'd like to test this
+behavior.
```
detiber: Should there be some warnings that:
- The external etcd cluster will be unmanaged and users will have to manage the lifecycle of those external etcd clusters appropriately
- The use of external etcd could incur additional data transit costs if deploying to a cloud provider depending on the provider configuration used

chuckha: Yes, better to provide these warnings now

docs/external-etcd.md
```diff
@@ -0,0 +1,77 @@
+# Support for external etcd
+
+Cluster API Bootstrap Provider Kubeadm  supports using an external etcd cluster for your workload Kubernetes clusters.
+
+### ⚠️ Warnings ⚠️
```
chuckha: @detiber Added this bit, please take a look!

docs/external-etcd.md
```diff
@@ -0,0 +1,77 @@
...
+
+Before getting started you should be aware of the expectations that come with using an external etcd cluster.
+
+* Cluster API is unable to manage any aspect of the external etcd cluster.
+* Depending on how you configure your etcd nodes you may incur additional cloud costs in data transfer.
+    * As an example, cross availability zone traffic can cost money on cloud providers. You don't have to deploy etcd
+    across availability zones, but if you do please be aware of the costs.
```
detiber: It applies to more than cross-AZ traffic with CAPA, especially if using the default VPC config, but I'm probably getting a bit too deep in the weeds for this doc :)

PR #273, opened 2019-10-04
🐛 Support running alongside other Cluster API pods in the same namespace with leader election enabled

### LINKED ISSUE #271: CABPK controller contending lock with CAPI
/kind bug

Running CABPK controller and CAPI controller in the same namespace. They are contending the leader elect lock

This is log from cabpk controller
```
yangshan-a02:v1alpha2 yangshan$ kubectl logs cabpk-controller-manager-644f88db48-s5npg -n b721480c4fc12e72896649c35c8543fb03ff1ea4 -f -c manager
I1004 16:29:14.967159       1 reflector.go:370] pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/reflector.go:94: Watch close - *v1alpha2.Cluster total 0 items received
I1004 16:29:16.789416       1 leaderelection.go:326] lock is held by capi-controller-manager-74cb5b985f-kfmt6_009aa8d2-e577-11e9-be08-629a49ac5bcb and has not yet expired
```

They should both aqquire their own leader elect lock

I am running a set up with CAPI+CAPA+CABPK in one namespace. CAPI and CAPA seem to live with each other just fine. From the log, there is no error

**Environment:**

- Cluster API version: v0.2.2
- Cluster API Bootstrap Provider Kubeadm version: 0.1.1
- Cluster API Infrastructure Provider version: v0.4.0
- Kubernetes version: (use `kubectl version`): 1.14
- OS (e.g. from `/etc/os-release`):

noamran:
Currently `LeaderElectionID` is set to the default `controller-leader-election-helper`. This causes an issue when several Cluster API pods are deployed in the same namespace. Adding a unique ID per service should resolve this. 
**Which issue(s) this PR fixes** :
Fixes #271

### DIFF
--- a/main.go
+++ b/main.go
@@ -103,6 +103,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+		LeaderElectionID:   "controller-leader-election-cabpk",
 		Namespace:          watchNamespace,
 		SyncPeriod:         &syncPeriod,
 	})


main.go
```diff
@@ -103,6 +103,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+		LeaderElectionID:   "controller-leader-election-helper-cabpk",
```
Lokicity: Non-blocking but it would be nice to make this into a constant

detiber:
```suggestion
		LeaderElectionID:   "controller-leader-election-cabpk",
```

I don't believe `helper` adds anything to the name of the configMap so it can likely be omitted, otherwise lgtm

detiber:
> Do we want to make this configurable in the future?

I don't believe so, since it is scoped to a namespace (We can set ctrl.Options.LeaderElectionNamespace in the case that `watchNamespace != ""`). Otherwise we are setting users up to deploy competing controllers.

ncdc:
Do you think it makes more sense to match the configmap's namespace to the pod's namespace or the watchNamespace? I guess I would expect it to match the pod.

detiber:
> Do you think it makes more sense to match the configmap's namespace to the pod's namespace or the watchNamespace? I guess I would expect it to match the pod.

Maybe...  This would be the default semantics today based on using the in cluster client for getting the Namespace.

detiber:
Using watchNamespace is nice, since we *know* the user is specifically limiting the watchNamespace to a single Namespace, but it also then possibly exposes the leader election configmap to users, which might not be desirable...  

That said, I wouldn't *expect* users to override the namespace for the Deployment unless they *knew* they wanted to run multiple controllers scoped down to a single namespace...

ncdc:
I would recommend (ordered by preference, high to low):

1. Match pod namespace, don't expose `--leader-election-namespace`
1. Match pod namespace, expose `--leader-election-namespace` to allow the user to customize
1. Match watchNamespace, don't expose `--leader-election-namespace`
1. Match watchNamespace, expose `--leader-election-namespace` to allow the user to customize

And 💯 that users 99.999% of the time don't need to know about the leader election namespace, or need to customize it.

detiber:
> Match pod namespace, don't expose --leader-election-namespace
> Match pod namespace, expose --leader-election-namespace to allow the user to customize

This exposes a similar footgun to overriding `leaderElectionID`

> Match watchNamespace, don't expose --leader-election-namespace
> Match watchNamespace, expose --leader-election-namespace to allow the user to customize

This exposes a similar footgun to overriding `leaderElectionID`

Another question I just thought of: Do we ever think it should be possible to deploy multiple CABPK( or any Cluster API controller) in the same namespace with different `watchNamespace`s? If so, maybe it would make sense to alter `leaderElectionID` to include the `watchNamespace` if set?

vincepri:
Should we move this conversation to an issue? I'd propose to merge this change as-is given that we're planning to release CABPK today, and explore different options (as outlined above) in a new issue, possibly making it consistent for all of CAP* projects.

How does that sound?

ncdc:
Yeah, I don't think we should expose any of this as flags.

I think including the watchNamespace in the name of the configmap could be a slippery slope - what if the watchNamespace's length is 250 characters (with the max being 253)? Then we can't have both `controller-leader-election-cabpk` and the watchNamespace in the name...

+1 to merging as-is (pending @detiber's suggestion to remove `helper`). Note, we'd have to think very carefully before we change the naming convention in this minor version (upgrade impact).

detiber:
So, the tricky part is that any change to the leaderElectionID and/or leaderElectionNamespace is a dangerous change...  The release note should include:

```
Before upgrading the controller deployment, first scale to 0, then apply the updated manifest, and then scale back to 1.
```

Otherwise there will be competing controllers using different resource locks during the course of the deployment.

I think we should be careful to limit these types of disruptive changes. If we are going to release today, then we should likely *not* include this change, *or* we should be good with the naming/namespacing convention we choose for at least a few releases.

ncdc:
@noamran FYI, the title of the PR is used in the release notes. I would recommend retitling to something like "Support running alongside other Cluster API pods in the same namespace with leader election enabled"

ncdc:
LGTM. Could you please squash down to 1 commit?

PR #275, opened 2019-10-07
🐛Change ownership of auto-generated certificate Secrets to the Cluster so they're not deleted alongside a Machine and KubeadmConfig

### LINKED ISSUE #274: Deleting first controlplane Machine deletes cert Secrets
/kind bug

- Create HA cluster, 1.15.3
- Run the extremely experimental upgrade tool to get to 1.16 ;)
- First controlplane node is excellent
- Second controlplane node never comes online
- Check namespace
- mfw certificate Secrets are missing 

~~A flawless upgrade, obviously!~~

I'm fairly certain this is because the cert Secrets have the first controlplane's KubeadmConfig as an owner reference. Deleting the first controlplane Machine deleted the KubeadmConfig, which in turn deleted the Secrets, which in turn made it so the 2nd new controlplane Machine had nothing to look up. This is a guess though.

/assign
/lifecycle active

rudoi:
This changes the owner of cert Secrets from the KubeadmConfig to the Cluster.

Fixes #274

### DIFF
--- a/internal/cluster/certificates.go
+++ b/internal/cluster/certificates.go
@@ -38,6 +38,7 @@ import (
 	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/kubeadm/v1beta1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha2"
 	"sigs.k8s.io/cluster-api/util/certs"
+	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/secret"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -194,6 +195,22 @@ func (c Certificates) Lookup(ctx context.Context, ctrlclient client.Client, clus
 			return err
 		}
 		certificate.KeyPair = kp
+
+		// Ensure Secret is owned by the Cluster. This provides a migration path for Secrets
+		// previously owned by KubeadmConfigs.
+		for _, ownerRef := range s.OwnerReferences {
+			if ownerRef.Kind == "KubeadmConfig" {
+				patchHelper, err := patch.NewHelper(s, ctrlclient)
+				if err != nil {
+					return err
+				}
+				s.OwnerReferences = certificate.AsSecret(cluster).OwnerReferences
+				if err := patchHelper.Patch(ctx, s); err != nil {
+					return err
+				}
+				break
+			}
+		}
 	}
 	return nil
 }
@@ -248,7 +265,7 @@ func (c Certificates) SaveGenerated(ctx context.Context, ctrlclient client.Clien
 		if !certificate.Generated {
 			continue
 		}
-		s := certificate.AsSecret(cluster, config)
+		s := certificate.AsSecret(cluster)
 		if err := ctrlclient.Create(ctx, s); err != nil {
 			return errors.WithStack(err)
 		}
@@ -304,7 +321,7 @@ func hashCert(certificate *x509.Certificate) string {
 }
 
 // AsSecret converts a single certificate into a Kubernetes secret.
-func (c *Certificate) AsSecret(cluster *clusterv1.Cluster, config *bootstrapv1.KubeadmConfig) *corev1.Secret {
+func (c *Certificate) AsSecret(cluster *clusterv1.Cluster) *corev1.Secret {
 	s := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: cluster.Namespace,
@@ -322,10 +339,10 @@ func (c *Certificate) AsSecret(cluster *clusterv1.Cluster, config *bootstrapv1.K
 	if c.Generated {
 		s.OwnerReferences = []metav1.OwnerReference{
 			{
-				APIVersion: bootstrapv1.GroupVersion.String(),
-				Kind:       "KubeadmConfig",
-				Name:       config.Name,
-				UID:        config.UID,
+				APIVersion: clusterv1.GroupVersion.String(),
+				Kind:       "Cluster",
+				Name:       cluster.Name,
+				UID:        cluster.UID,
 			},
 		}
 	}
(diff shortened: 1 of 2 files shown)


internal/cluster/certificates.go
```diff
@@ -322,10 +322,10 @@ func (c *Certificate) AsSecret(cluster *clusterv1.Cluster, config *bootstrapv1.K
 	if c.Generated {
 		s.OwnerReferences = []metav1.OwnerReference{
 			{
-				APIVersion: bootstrapv1.GroupVersion.String(),
-				Kind:       "KubeadmConfig",
-				Name:       config.Name,
-				UID:        config.UID,
+				APIVersion: clusterv1.GroupVersion.String(),
+				Kind:       "Cluster",
```
rudoi: sidenote: this hardcode feels weird. nowhere better to get this?

rudoi:
> I don't think this func needs to take in `config` as a param any more?

oooo nice catch! removing params is my favorite.

ncdc:
I'd also recommend retitling this PR to something like
> Change ownership of auto-generated certificate secrets to the Cluster so they're not deleted alongside a Machine and KubeadmConfig

rudoi:
@ncdc sounds good to me, provided you don't mind my blatant copypasta

ncdc:
/approve

Do we need to do anything in the code to deal with preexisting clusters that have the ownerRefs set to the KubeadmConfig?

detiber:
Would it make sense to handle updating of ownerRefs for secrets that currently list the current KubeadmConfig as an owner? Otherwise we should make sure that the release notes include a migration path for existing Clusters.

rudoi:
> Would it make sense to handle updating of ownerRefs for secrets that currently list the current KubeadmConfig as an owner? Otherwise we should make sure that the release notes include a migration path for existing Clusters.

I think it'd be good to do this in code. Separate issue perhaps?

ncdc:
It'd be nicer to do it with this PR if you can?

rudoi:
-sigh- anything for y'all! gimme a few 😄

internal/cluster/certificates.go
```diff
@@ -194,6 +195,17 @@ func (c Certificates) Lookup(ctx context.Context, ctrlclient client.Client, clus
...
 		certificate.KeyPair = kp
+
+		// Ensure Secret is owned by the Cluster. This provides a migration path for Secrets
+		// previously owned by KubeadmConfigs.
+		patchHelper, err := patch.NewHelper(s, ctrlclient)
+		if err != nil {
+			return err
+		}
+		s.OwnerReferences = certificate.AsSecret(cluster).OwnerReferences
+		if err := patchHelper.Patch(ctx, s); err != nil {
```
rudoi: meh I don't like what I did here. this only updates the ownerReferences on a new join, so conceivably you could have this code deployed and still cascade delete it by deleting the Machine. Should there just be a lookup as part of the core reconciliation path?

internal/cluster/certificates.go
```diff
@@ -194,6 +195,17 @@ func (c Certificates) Lookup(ctx context.Context, ctrlclient client.Client, clus
...
 		}
 		certificate.KeyPair = kp
+
+		// Ensure Secret is owned by the Cluster. This provides a migration path for Secrets
+		// previously owned by KubeadmConfigs.
+		patchHelper, err := patch.NewHelper(s, ctrlclient)
+		if err != nil {
+			return err
+		}
+		s.OwnerReferences = certificate.AsSecret(cluster).OwnerReferences
```
detiber: This should probably only override the OwnerReferences if the OwnerRef is currently a `KubeadmConfig`, otherwise we might take ownership of a Secret we shouldn't.

rudoi: ah right, forgot about BYOCerts

chuckha:
/hold

I don't believe certs being owned by the cluster is the correct path here.

There is a fix in amy's branch that addresses the secret deletions by updating the owning kubeaconfig reference: https://github.com/vmware/cluster-api-upgrade-tool/pull/72/files#diff-4410eeb26984da2b31d9a58c8a250133R361

detiber:
> There is a fix in amy's branch that addresses the secret deletions by updating the owning kubeaconfig reference: https://github.com/vmware/cluster-api-upgrade-tool/pull/72/files#diff-4410eeb26984da2b31d9a58c8a250133R361

One worry I have with this approach is that it only solves the upgrade use case, replacing a failed control plane instance would need similar manual workarounds.

imo, control plane management should help out in this area for v1alpha3+ at least.

chuckha:
One usecase this does not address is updating serving cert SANs. If the certs are owned by the cluster we have to delete the cluster to get the secrets to delete or delete them by hand. If they are owned by the kubeadmconfig then we can scale the control plane machines down to 0, modify the kubeadmconfig and bring up a control plane machine to generate new updated certificates.

internal/cluster/certificates.go
```diff
@@ -194,6 +195,22 @@ func (c Certificates) Lookup(ctx context.Context, ctrlclient client.Client, clus
 			return err
 		}
 		certificate.KeyPair = kp
+
+		// Ensure Secret is owned by the Cluster. This provides a migration path for Secrets
+		// previously owned by KubeadmConfigs.
+		for _, ownerRef := range s.OwnerReferences {
+			if ownerRef.Kind == "KubeadmConfig" {
```
vincepri:
```suggestion
			if ownerRef.Kind != "Cluster" {
```

detiber: @vincepri we just check to see if it is not owned by a Cluster and someone is managing the secret externally and setting their own OwnerRef on it, then we would take ownership of it erroneously.

chuckha:
@rudoi after our discussion yesterday do you still feel this is a necessary change?

rudoi:
> @rudoi after our discussion yesterday do you still feel this is a necessary change?

No, not at this time. I think there's some theorycrafting to be done regarding cert ownership outside of the upgrade flow, but we can probably defer on that.

PR #278, opened 2019-10-11
⚠️ Use controller-runtime client everywhere for consistency 

### LINKED ISSUE #127: Use controller-runtime client everywhere for consistency
/kind feature

[A clear and concise description of what you want to happen.]
In the reconcile logic in the tests we have both static and controller-runtime clients in play. We should standardize on the controller-runtime client, as it is awesome. 

Basically the PRs for the bootstrap token logic and the cert generation logic were being worked at the same time, so unfortunately we ended up with two different methods for creating/getting secrets, so that should be reconciled as well.

tahsinrahman:

Fixes #127

### DIFF
--- a/controllers/kubeadmconfig_controller.go
+++ b/controllers/kubeadmconfig_controller.go
@@ -26,7 +26,6 @@ import (
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


ncdc:
/hold

We are prepping to move the code in this repo to the cluster-api repo. Would you be willing to recreate this against cluster-api after the code has moved?

https://github.com/kubernetes-sigs/cluster-api/pull/1483

tahsinrahman:
Sure! 
/close

PR #287, opened 2019-12-17
⚠️ ClusterConfigration.Networking.* fields are now optional

vincepri:

Fixes a long standing issue that was causing pain and fear 😂 

/assign @chuckha @ncdc

### DIFF
--- a/kubeadm/v1beta1/types.go
+++ b/kubeadm/v1beta1/types.go
@@ -227,11 +227,13 @@ type NodeRegistrationOptions struct {
 // Networking contains elements describing cluster's networking configuration
 type Networking struct {
 	// ServiceSubnet is the subnet used by k8s services. Defaults to "10.96.0.0/12".
-	ServiceSubnet string `json:"serviceSubnet"`
+	// +optional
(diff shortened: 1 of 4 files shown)


kubeadm/v1beta1/types.go
```diff
@@ -227,11 +227,14 @@ type NodeRegistrationOptions struct {
 // Networking contains elements describing cluster's networking configuration
 type Networking struct {
 	// ServiceSubnet is the subnet used by k8s services. Defaults to "10.96.0.0/12".
-	ServiceSubnet string `json:"serviceSubnet"`
+	// +optional
+	ServiceSubnet string `json:"serviceSubnet,omitempty"`
 	// PodSubnet is the subnet used by pods.
-	PodSubnet string `json:"podSubnet"`
+	// +optional
```
ncdc: It looks like kubeadm will default ServiceSubnet and DNSDomain but not PodSubnet. Should we keep this one as required?

vincepri: Yeah, +1

PR #289, opened 2020-01-17
✨ import and enable pprof

noamran:
Following up on this [PR](https://github.com/kubernetes-sigs/cluster-api/pull/2084). Adding `pprof` to this repo.


chuckha:
/ok-to-test

If you've got ~10 commits here you could apply to the kubernetes-sigs organization so you don't need to have someone add the /ok-to-test label!

chuckha:
/lgtm

just realized which repo we were in, but this seems like a fine thing to back port

